PR #35, opened 2021-01-29
#23-Adding AWS info as database provider to readme

### LINKED ISSUE #23: Research - Database Hosting Platform
Explore PostgreSQL hosting providers in an effort to understand which cloud-based fully-managed database hosting provider would be best. Also see #4. Options include: (feel free to trim the list before digging too deep)

- [Digital Ocean](https://www.digitalocean.com/products/managed-databases-postgresql/)
- [Azure PostgreSQL](https://azure.microsoft.com/en-us/services/postgresql/)
- [AWS RDS for PostgreSQL](https://aws.amazon.com/rds/postgresql/)
- [Google](https://cloud.google.com/sql/docs/postgres)
- [ScaleGrid](https://scalegrid.io/pricing.html?db=POSTGRESQL&cloud=cloud_oci&replica=deployment_standalone&instance=Medium#section_pricing_dedicated)
- [Heroku PostgreSQL](https://www.heroku.com/postgres)
- [ElephantSQL](https://www.elephantsql.com)

Pricing is likely the most important axis of comparison. From rough napkin math, we anticipate having <1GB of data and a relatively small CPU load, but this could change. The 1-3 lowest price tiers are likely of most interest, and it would be beneficial to compare and see how features / performance scales with price for each provider. If prompted, bias towards hosting on the east coast, but do consider the price differences.

eshwaribhide:
Closes #23 
Also, James is fantastic

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,6 @@ The below table will help track various components of setting up this project an
 | Styling | Bootstrap | [Y](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/19) |  |  |  |
 | URL Routing | React Router | [Y](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/21) |  |  |  |
 | Continuous Integration | GitHub Actions | [Y](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/22) |  |  |  |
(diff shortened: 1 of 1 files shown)


jamescd18 (approved):
Looks wonderful! Glad to have this decided so we can move forward.

Issue #2, opened 2021-01-20
Research - Prisma as an Object Relational Mapper


jamescd18:
Research [Prisma](https://www.prisma.io) for use as an [Object Relational Mapper (ORM)](https://blog.bitsrc.io/what-is-an-orm-and-why-you-should-use-it-b2b6f75f5e2a). Potentially consider [alternatives](https://www.prisma.io/dataguide/database-tools/evaluating-type-safety-in-the-top-8-typescript-orms) like [TypeORM](https://typeorm.io/#/), [Sequelize](https://sequelize.org), or [Objection.js](https://vincit.github.io/objection.js/) to achieve the same goals. Consider [benefits](https://www.reddit.com/r/node/comments/l00fit/what_do_you_think_about_orms/), ease of modifying the schema, and compatibility with PostgreSQL and TypeScript.

Put longer notes here in this issue, along with potential sample code and any other thoughts. And then you can put a couple bullets summary / thoughts in [this document](https://docs.google.com/document/d/1hKDfqYFajrCg6U0QhOf75hAwOhQ7082yRn5ZMaoqE4g/edit#).

See Northeastern-Electric-Racing/NER-PM-Dashboard#114 as an excellent and thorough example.

jamescd18:
So far it looks like Prisma would be a pretty great option. Here's some resources I found while digging:
- [a tutorial for deploying to Netlify with Prisma integrated](https://www.prisma.io/docs/guides/deployment/deploying-to-netlify)
- [an article about setting up prisma + PostgreSQL from scratch](https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch-typescript-postgres)

Looks like it should be super simple

eshwaribhide:
Can this be closed?

PR #39, opened 2021-01-30
Decided on Prisma for ORM

jamescd18:
After doing a bit of research detailed in #2, I think it's safe to decide on Prisma for the ORM.


kevinyu328 (approved):
Prisma it is!

PR #41, opened 2021-01-30
Set up prisma & local PostgreSQL

jamescd18:
- Installed Prisma items
- added Prisma migration command to scripts
- added PostgreSQL install to docs
- tested Prisma with example schema, app running locally, localhost database


kevinyu328 (dismissed):
Confused about some things, but looks good to me! @eshwaribhide's review will probably be more useful than mine

prisma/migrations/20210130095228_add_user_age/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "User" ADD COLUMN     "age" INTEGER;
```
eshwaribhide, line 2: The extra space bothers me

Docs/Onboarding.md
```diff
@@ -1,11 +1,30 @@
...
+`DATABASE_URL="postgresql://<USERNAME>:@localhost:5432/nerpm?schema=public"`
+
+To test that things are working, run `npm run start` in the CLI and go to an example API route.
+If the starter code still exists, use `localhost:3000/.netlify/functions/seed`.
+
+## IDE
+
+These are extensions for VSCode that might make your developer experience more enjoyable.
+They are completely optional.
+
```
eshwaribhide, line 29: This is not related to this PR, but I believe there's a VSCode extension for Prettier too, I know we're implementing it for our repo but maybe you could still add the vscode ext info here

prisma/migrations/20210130094615_init/migration.sql
```diff
@@ -0,0 +1,42 @@
+-- CreateTable
+CREATE TABLE "Post" (
```
eshwaribhide, line 2: Maybe I'm mistaken, but are you sure you need quote marks around the table name & column names?

eshwaribhide: In the sense that, I think it's fine but I don't think you need them unless you need to escape certain things (e.g. a column named "serial" to not confuse its name with the data type)

prisma/migrations/20210130094615_init/migration.sql
```diff
@@ -0,0 +1,42 @@
...
+    "userId" INTEGER NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "User" (
+    "id" SERIAL NOT NULL,
+    "email" TEXT NOT NULL,
+    "name" TEXT,
```
eshwaribhide, line 27: Shouldn't we also add a "NOT NULL" constraint to name?

prisma/schema.prisma
```diff
@@ -0,0 +1,38 @@
...
+  id     Int     @default(autoincrement()) @id
+  bio    String?
+  user   User    @relation(fields: [userId], references: [id])
+  userId Int     @unique
+}
+
+model User {
+  id      Int      @default(autoincrement()) @id
+  email   String   @unique
+  name    String?
```
eshwaribhide, line 34: In the sql file had said maybe name should be not null, so if we decide to implement that then also make name not optional here

prisma/schema.prisma
```diff
@@ -0,0 +1,38 @@
...
+model Post {
+  id        Int      @default(autoincrement()) @id
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+  title     String
+  content   String?
+  published Boolean  @default(false)
+  author    User     @relation(fields: [authorId], references: [id])
+  authorId  Int
+}
```
eshwaribhide, line 22: Shouldn't authorId also have "@unique" next to it?

Docs/Onboarding.md
jamescd18: yeah, I was gonna add those as part of other PRs

jamescd18:
The prisma schema is just some dummy data, and Prisma auto-generates all the SQL stuff, so I think we should be good

eshwaribhide (approved):

PR #82, opened 2021-02-23
Decided on URL Routing

jamescd18:
Updating docs to signify we've decided to use react router for the url routing


kevinyu328 (approved):
nice!

PR #93, opened 2021-02-27
Created Projects Detail Page to Display WBS

jamescd18:
Created a new page that will display either project details or work package details when given the WBS number. Reconfigured various different shared code, added WBS number validation from v1 repo, and the routing wiring to accommodate this new change. This change results in being able to click on rows in the projects table and be navigated to a page to see the details for the desired WBS number.

Projects table:
Screen Shot 2021-02-27 at 1 51 14 AM

Navigated via click on row:
Screen Shot 2021-02-27 at 1 51 25 AM

Navigated via direct URL navigation:
Screen Shot 2021-02-27 at 1 52 28 AM

### DIFF
--- a/src/pages/projects/projects.tsx
+++ b/src/pages/projects/projects.tsx
@@ -3,14 +3,19 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
+import { Route, Switch } from 'react-router-dom';
 import ProjectsTable from '../../containers/projects-table/projects-table';
+import WBSDetails from '../wbs-details/wbs-details';
 import './projects.module.css';
 
 const Projects: React.FC = () => {
   return (
     <div>
       <h1>This is the Projects Page</h1>
-      <ProjectsTable />
+      <Switch>
+        <Route path="/projects/:wbsNum" component={WBSDetails} />
+        <Route path="/projects" component={ProjectsTable} />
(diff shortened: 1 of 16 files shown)


src/pages/projects/projects.tsx
```diff
@@ -3,14 +3,19 @@
...
 import './projects.module.css';
 
 const Projects: React.FC = () => {
   return (
     <div>
       <h1>This is the Projects Page</h1>
-      <ProjectsTable />
+      <Switch>
+        <Route path="/projects/:wbsNum" component={WBSDetails} />
+        <Route path="/projects" component={ProjectsTable} />
```
kevinyu328, line 17: Will this route be in conflict with the route in App.tsx? Or is it because that this route is rendered within projects.tsx, so it only routes when both the path is /projects and the component that is currently rendered is projects.tsx?

kevinyu328 (commented):
One small question

src/pages/projects/projects.tsx
jamescd18: I don't think there's any conflict, and I think they do it the same way in their example: https://reactrouter.com/web/example/nesting

kevinyu328 (approved):
Whoops sorry, should've been merged a while ago.

Issue #60, opened 2021-02-04
General - Create Sidebar Component 


kevinyu328:
Create the sidebar component for the website. The plan is to have each icon as an individual component and to have one component that includes all the individual icon components. Refer to the V3 design sketch for how the sidebar should look. @jamescd18 has some Font Awesome icons in mind, so reach out to him for the specifics. The background color of the sidebar is undecided, so feel free to play around with colors to see what you think looks best. You can create separate files to show different styles/colors or take screenshots of some of the different styles/colors you experimented with. [This document ](https://docs.google.com/document/d/1bV-TGorNeGbn7BhrwP9_nhSb_qmfs5wxJeJY5AfekDo/edit) has some of the specific color codes the club uses for branding.

jamescd18:
Here is a sneak peak at design sketch v4. Icons are pulled from font awesome. To be extra specific:

- generic home icon
- bar chart turned on its side
- folder icon
- two arrows pointing opposite directions left and right
- stonks line chart going up to the right
- question mark help icon

![image](https://user-images.githubusercontent.com/51571627/106844560-b6cfde80-665d-11eb-8a11-eb8db511c24f.jpeg)

albeartoz:
To consider:
 - https://www.npmjs.com/package/react-pro-sidebar 
 - https://www.npmjs.com/package/react-sidebar 
 - https://www.npmjs.com/package/react-burger-menu

jamescd18:
Awesome! Just like you said, best not to reinvent the wheel. Looking a little more closely at these three, it seems like either way, it'll require a bit of customization to get it to where we'd like it to be. The first is a lot newer and less popular than the other two, but that shouldn't matter too much. Just pick whichever you think would be best, and style it appropriately to align with the vision outlined in the sketch picture above.

jamescd18:
This is what the new sidebar should look like
<img width="1079" alt="Screen Shot 2022-01-20 at 1 09 28 PM" src="https://user-images.githubusercontent.com/51571627/150396751-11da6b2d-e0b4-4e62-bc71-1a8164592714.png">

PR #104, opened 2021-03-17
#60 albert: Sidebar

albeartoz:
I screwed up some of my commits to not tag the correct issue, but I hope that's ok. I'm not sure how to go back and fix it at this point. Otherwise, the sidebar is done, and supports React Router when we'll need it. There's a slight CSS hiccup in that the sidebar does not extend down the entire side, but other than that, it's done.


jamescd18 (changes requested):
Thank you Albert! Really great start, and I love getting to see this PR! I've summarized the action items below
- [ ] Clean up linting errors
- [ ] Integrate with the router
- [ ] Wrap test method in a test suite `describe()`
- [ ] Split single test case into a test method for each button (this will make it so that a change to a single button will only cause at most 1 test to fail, and ensure the rest do not also fail for some reason)

jamescd18:
Stale PR, closing.

PR #107, opened 2021-03-19
#68 - Added Dummy API Endpoint for Change Requests

### LINKED ISSUE #68: Change Requests - Create Dummy CR API Endpoint
Create a dummy API endpoint similar to #67 for the `/change-requests` path. Returned data should be an array of change requests. Change requests have at bare minimum an ID, associated WBS#, submitter, and type. 

---

Here's a quick outline of potential steps for this issue.
- [x] Add file to `/backend/functions/`
- [x] Add test file to `/backend/tests/`
- [x] Copy in basic serverless endpoint handler function
- [x] Add dummy data hard-coded into the file (see #74 for example)
- [ ] Add proper interface type
- [ ] Configure the handler to supply a work package as the return
- [ ] Add basic `describe()` test suite to test file
- [ ] Add a couple `it()` test cases to test suite

Not sure if this will work properly for someone else working on the issue, but it's good to have here either way.

TheElbowTurnMaster:
closed #68 
I learned more about REST and aws-lambda while looking at this, but I'm not sure what else needs to be here other than everything I copied from the project function. I assume it's just a GET request, but I would like to know what I missed.


src/backend/functions/change-requests.ts
```diff
@@ -0,0 +1,37 @@
+import { Context } from 'aws-lambda';
+import { ChangeRequest } from 'utils/src/types/change-request-types';
```
jamescd18, line 2: Your current import statement here is a file path reference, but in order to properly have the deployment automation bundle all the needed files for the AWS lambda build, it needs to be a package reference instead. 

Basically it needs to just be `'util'`, but if you don't understand the above statement, let me know and I'll send you some resources to learn about stuff.

src/backend/functions/change-requests.ts
```diff
@@ -0,0 +1,37 @@
...
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(exampleCRs)
+    };
+  } catch (error) {
+    console.error(error);
+    return { statusCode: 500 };
+  }
+}
+
+const changeRequest1: ChangeRequest = {
```
jamescd18, line 23: These fields don't look correct, are you sure the type-checking is happy with it? If yes, we'll definitely need to wait until my massive PR gets merged before merging this one too.

src/backend/tests/change-requests.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      expect(cr.hasOwnProperty('id')).toBeTruthy();
+      expect(cr.hasOwnProperty('submitter')).toBeTruthy();
+      expect(cr.hasOwnProperty('type')).toBeTruthy();
+    });
+  });
+
+  it('has proper project wbsNums', () => {
+    exampleCRs.forEach((cr) => {
+      const changeRequest: ChangeRequest = cr;
+      const wbs: Array<string> = changeRequest.wbsNum.split('.');
```
jamescd18, line 21: lint error

src/utils/src/types/change-request-types.ts
```diff
@@ -4,6 +4,8 @@
  */
 
 export interface ChangeRequest {
-  submitter: String;
-  type: String;
+  id: number;
```
jamescd18, line 7: what's the reason for changing the data definition? generally data definition changes are more critical and have the potential to break things, so unless explicitly detailed in the issue, you'll want to avoid data definition changes.

jamescd18 (changes requested):
Great start! I think it'd be really beneficial for you to ensure you understand the why behind all of this code. I'd highly suggest checking out the v1 application on the reports page, select "all change requests" and explore the various data that gets returned.

jamescd18:
Now that the massive PR has been merged, make sure to update and fast forward your branch. Then examine the data types and new location for dummy data, and refactor / update / improve the code for this PR so it integrates nicely. Let me know if you run into any issues!

src/backend/functions/change-requests.ts
jamescd18: These should be swapped out for the ones from the `dummy-data.ts` file from the `utils` package

jamescd18 (approved):
Fantastic job! This is a great implementation for the change requests API matching the exiting design pattern.

PR #122, opened 2021-03-28
Updated Users API Endpoint

jamescd18:
I've updated the users endpoint to now return just the dummy data from the utils package. Added tests for the endpoint handler


kevinyu328 (approved):
Nice nice

PR #123, opened 2021-03-28
Create Login Page

### LINKED ISSUE #120: Auth - Create a Basic Login Page
In order to facilitate having users logging into the application, we need a login screen. Since we'll be using Google as our primary authentication provider, this login page won't be like traditional login pages that ask for username and password. Instead, it'll just have a button to trigger Google login. That said, we'll need the back-end set up in order to handle that, so for now, we'll just have a button that sends the user to the home page.

One interesting thing about this page is that it shouldn't have any of the header or nav bar like the rest of the application. This is because we want users to understand that they can't use the application without logging in. See the design sketch below for details.

Screen Shot 2021-03-26 at 10 44 02 PM

jamescd18:
Created the login component and temporarily wired it in with the rest of the app on the `/login` url. Currently the login button just takes you to the home page. Closes #120 
NER_PM_Login

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,6 +8,7 @@ import ChangeRequests from './pages/change-requests/change-requests';
 import Projects from './pages/projects/projects';
 import './App.css';
 import { PageNotFound } from './pages/page-not-found/page-not-found';
+import Login from './pages/login/login';
 
 function App() {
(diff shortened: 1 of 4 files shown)


kevinyu328 (approved):
Nice

PR #127, opened 2021-04-08
i91 Logan - Create a Work Packages Summary Component

### LINKED ISSUE #91: Projects - Create a Work Packages Summary Component
Create a component that will be able to display a summary of a work package. Display the summary according to the design sketch below. I would imagine there to be the `WorkPackagesAccordion` component which displays [an accordion style list](https://getbootstrap.com/docs/4.2/components/collapse/#accordion-example) with extra details within each collapsable row. This is a presentational / view-only component that takes in props and displays them.

lpfahler16:
I have implemented all of the logic and style for this issue and all I have left is tests. Thought it might be useful to get feedback before spending too much time on tests. Specifically, what should the number on the right side of the header of the card be? Right now it is the amount of time from today's date (though I inputed an earlier for testing) to the end date. Is that what it should be or just the duration. Testing should be finished this weekend. Closes #91


src/components/work-package-summary/work-package-summary.tsx
```diff
@@ -0,0 +1,128 @@
...
+      return diffDays + " days";
+    }
+    
+  }
+};
+
+// Formats the WBS numbers
+const formatWbsNum: Function = (wbs: WbsNumber): string => {
+  return wbs.area + "." + wbs.project + "." + wbs.workPackage;
+};
```
kevinyu328, line 53: Can use wbsPipe instead

src/components/work-package-summary/work-package-summary.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+      
+      <Collapse in={open}>
+          <div>
+            <Card.Body>
+              <p>{workPackage.deliverable}</p>
+              <div className={styles.halfDiv}>
+                <p><b>Dependencies:</b> {formatDependencies(workPackage.dependencies)}</p>
+                <p><b>Rules:</b> {formatRules(workPackage.rules)}</p>
+                <p><b>Budget:</b> ${workPackage.budget}</p>
```
kevinyu328, line 115: Can use dollarsPipe

src/components/work-package-summary/work-package-summary.tsx
```diff
@@ -0,0 +1,128 @@
...
+    return 0 + " days";
+
+  } else if (diffDays > 7) {
+
+    var diffWeeks = diffDays / 7;
+
+    if (diffWeeks === 1) {
+      return diffWeeks + " week";
+    } else {
+      return diffWeeks + " weeks";
```
kevinyu328, line 36: Can use weeksPipe

kevinyu328 (changes requested):
Amazing work Logan! Thank you for taking the time to build this component! I think the number on the right side of the card header is the duration. @jamescd18 can you confirm?Just a couple small things that I noticed as well:

1. There is a pipes.tsx file under the shared folder that abstracts some of the view formatting that you had to do to create the component. Those functions should make formatting easier, so take a look at those functions and see where they can be reused. 
2. From what I understand, the unit of measurement used for project management within the club is in weeks, so I don't think we will need to display the days. The smallest unit of measurement would be one week, if that makes sense.

Again, great work!

kevinyu328 (commented):
Nice, looks good to me! Just need @jamescd18 to confirm that the number on the right side of the card is indeed duration and not the remaining weeks.

lpfahler16:
Awesome. Just have to finish up testing too but I'll let you know when that is done. I'd love to work on a similar issue next, (possibly 72) but something completely different would be fine too.

jamescd18:
Can you also attach a screenshot of the completed component? This is really helpful for us reviewing front-end pull requests.

lpfahler16:
Screen Shot 2021-04-10 at 4 13 44 PM
Screen Shot 2021-04-10 at 4 13 32 PM

All set with testing and ready to merge on my end. Attached are images of the component. I did not include any margins around them in the final code. Rather, I added an optional className prop so the component can be styled when it is used (similar to a bootstrap component).

src/components/work-package-summary/work-package-summary.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <p>{workPackage.deliverable}</p>
+            <div className={styles.halfDiv}>
+              <p>
+                <b>Dependencies:</b> {formatList(workPackage.dependencies, wbsPipe)}
+              </p>
+              <p>
+                <b>Rules:</b>{' '}
+                {formatList(workPackage.rules, (str: string): string => {
+                  return str;
+                })}
```
jamescd18, line 69: use an implicit return to shorten the lambda?

jamescd18 (changes requested):
This is amazing work! A really well put together component that demonstrates clean usage of react and bootstrap with pleasing design layout.

A few different things I've outlined in inline review comments:
- The new pipes you created are perfect! Let's move them to the shared folder where pipes live so we can have cleaner separation and these pipes can be used for other components too
- I've added a few suggestions on name changes in an effort to help ensure our vocabulary is consistent. Aligning the naming of things within the code to project management naming can hopefully reduce confusion.
- I've suggested some simplifications in a few places, but feel free to reply and let me know if things don't work the way I was thinking / hoping for.
- A lot of the suggestion are made with the collaborative and long-running nature of this project. A lot of different people are likely to work on it over many years, so we'll want to ensure code is clear and easily understandable. (Just like in industry)

Really awesome job here. Let me know if anything doesn't make sense or if you're curious about why for any of my review comments. 👍

src/components/work-package-summary/work-package-summary.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    expect(screen.getByText(`${wp.deliverable}`)).toBeInTheDocument();
+    expect(
+      screen.getByText(`Dependencies:${listPipe(wp.dependencies, wbsPipe)}`)
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        `${listPipe(wp.rules, (str: string): string => {
+          return str;
+        })}`
```
jamescd18, line 27:
```suggestion
        `${listPipe(wp.rules, (str: string) => str)}`
```
Nice spot to use implicit return. I believe you can also remove the explicit return type because you're just returning the parameter so the type system can infer the return type really easily. I would recommend using this method to shorten the identity function in the rest of your files since it's so simple.

src/components/work-package-summary/work-package-summary.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    expect(screen.getByText(`${wbsPipe(wp.wbsNum)}`)).toBeInTheDocument();
+    expect(screen.getByText(`${wp.duration} weeks`)).toBeInTheDocument();
+
+    expect(screen.getByText(`${wp.deliverable}`)).toBeInTheDocument();
+    expect(screen.getByText(`${listPipe(wp.dependencies, wbsPipe)}`)).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        `${listPipe(wp.rules, (str: string): string => {
+          return str;
+        })}`
```
jamescd18, line 69: Similarly shorten with implicit return?

src/shared/pipes.tsx
```diff
@@ -36,3 +36,16 @@ export const fullNamePipe: Function = (user: User): string => {
 export const booleanPipe: Function = (bool: boolean): string => {
   return bool ? 'YES' : 'NO';
 };
+
+// Formats an array of objects into a string that is a list
+export const listPipe: Function = <T,>(array: T[], transform: (ele: T) => string): string => {
+  var strArray = array.map(transform);
+  return strArray.join(', ');
```
jamescd18, line 43:
```suggestion
  return array.map(transform).join(', ');
```
Pretty sure this can be shortened with chaining since `array.map()` returns a new array.

src/components/work-package-summary/work-package-summary.test.tsx
```diff
@@ -0,0 +1,76 @@
...
```
jamescd18, line 48: Shorten with implicit return

src/components/work-package-summary/work-package-summary.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <Collapse in={open}>
+        <div>
+          <Card.Body>
+            <p>{workPackage.deliverable}</p>
+            <div className={styles.halfDiv}>
+              <p>
+                <b>Dependencies:</b> {listPipe(workPackage.dependencies, wbsPipe)}
+              </p>
+              <p>
+                <b>Rules:</b> {listPipe(workPackage.rules, (str: string): string => str)}
```
jamescd18, line 51: infer return type?

jamescd18 (changes requested):
This looks fantastic! You did a great job of cleaning up everything I noted. Just a few nitpick items I've noted. It's useful to understand some of the small nuances of JS/TS with lambda implicit returns and the TS type system's inferred types to reduce code and increase readability while not compromising on understandability.

jamescd18 (approved):
Amazing work, and really great follow-through on the reviews! I think it's all good to go.

kevinyu328 (approved):
Wow, looks good! Thanks for the detailed review James!

PR #132, opened 2021-04-11
Built API Route Matching, Integrated with projects endpoint

### LINKED ISSUE #109: Projects - Wrong statusCode in API
`statusCode: 201,` this should be 200 (Ok) instead of 201 (created) since the endpoint is not creating anything

Observed Behavior: Projects API endpoint returns status code 201
Expected Behavior: Projects API endpoint should return status code 200
Steps to Reproduce: Query the projects API endpoint 
Screenshots: n/a

### LINKED ISSUE #117: General - Devise a Method for Sub-Route Matching for API Endpoints
Because we're using Netlify functions and AWS lambdas, each file in the API back-end gets assigned to an api endpoint named after the file. In order to have meaningful API design, we should be defining sub-routes for each specific method in the end-point. In order to do this, we need to be able to have path matching to easily associate a path and http method to a function.

We'll use the [`path-to-regexp`](https://www.npmjs.com/package/path-to-regexp) package to help convert path definitions and confirm against endpoint query paths. This can potentially be done by defining an API route object which contains the path, http method, and function to call, and store an array of those objects. Then potentially whenever a query comes in, iterate over the routes until a matching route + http method combo is found, then invoke the function and pass it the desired fields.

jamescd18:
Closes #117. Closes #109.


src/backend/tests/projects.test.ts
```diff
@@ -3,27 +3,78 @@
...
+      expect(projectResponse.hasOwnProperty('wbsNum')).toBeTruthy();
+      expect(projectResponse.hasOwnProperty('name')).toBeTruthy();
+      expect(projectResponse.hasOwnProperty('projectLead')).toBeTruthy();
+      expect(projectResponse.hasOwnProperty('projectManager')).toBeTruthy();
+    });
+
+    it('has proper project wbsNums', () => {
+      expect(projectResponse.wbsNum).toBeTruthy();
+      expect(projectResponse.wbsNum.workPackage).toEqual(0);
     });
```
kevinyu328, line 78: Just to be thorough, adding a test case for when the wbs specified in the path does not exist could be helpful

kevinyu328 (changes requested):
Would just add one more test case to be thorough. Otherwise, it looks good to me

kevinyu328 (dismissed):
Nice!!

kevinyu328 (approved):
Nice

Issue #111, opened 2021-03-24
General - Centralize Back-End Routes in the Utils Package


jamescd18:
In order to reduce duplication and errors, and to make things easier to find, we should consolidate the definition of back-end API routes in a single object defined in the utils package so it is available to all API endpoints.

jamescd18:
Saving this for reference: https://forbeslindesay.github.io/express-route-tester/

PR #133, opened 2021-04-11
Added central api routes file

jamescd18:
Added a file to the utils package for use in centralizing route definition for the api routes.

### DIFF
--- a/src/utils/src/api-routes.ts
+++ b/src/utils/src/api-routes.ts
@@ -1,0 +1,36 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+/**************** Users Endpoint ****************/
+const USERS: string = `/users`;
+const USERS_LOGIN: string = `${USERS}/login`;
+
+/**************** Projects Endpoint ****************/
+const PROJECTS: string = `/projects`;
+const PROJECTS_SINGLE: string = `${PROJECTS}/:wbs`;
+
+/**************** Work Packages Endpoint ****************/
+const WORK_PACKAGES: string = `/work-packages`;
+const WORK_PACKAGES_SINGLE: string = `${WORK_PACKAGES}/:wbs`;
+
+/**************** Change Requests Endpoint ****************/
+const CHANGE_REQUESTS: string = `/projects`;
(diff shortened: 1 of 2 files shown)


jamescd18:
Closes #111

src/utils/src/api-routes.ts
```diff
@@ -0,0 +1,36 @@
...
+const PROJECTS_BY_WBS: string = `${PROJECTS}/:wbs`;
+
+/**************** Work Packages Endpoint ****************/
+const WORK_PACKAGES: string = `/work-packages`;
+const WORK_PACKAGES_BY_WBS: string = `${WORK_PACKAGES}/:wbs`;
+
+/**************** Change Requests Endpoint ****************/
+const CHANGE_REQUESTS: string = `/change-requests`;
+const CHANGE_REQUESTS_BY_ID: string = `${CHANGE_REQUESTS}/:id`;
+const CHANGE_REQUESTS_REVIEW: string = `${CHANGE_REQUESTS_BY_ID}\\:review`;
```
kevinyu328, line 21: What's the double backslash used for?

jamescd18: The first backslash acts as a string escape, and the second one allows me to type the backslash. If it were just one backslash, it would escape the next character. I want to keep the backslash in order to have the string subsequently be able to escape the colon when the string gets evaluated by the route matcher.

If you check out [this Express Route Tester](https://forbeslindesay.github.io/express-route-tester/), select package version 2.0.0 (I believe express and react-router use 1.7.0, I have no idea why, but the behavior I want is in 2.0.0 or newer versions so I'm using 6.x.x or something in the utils package). Plug in `/change-requests/:id\:review` for the route and `/change-requests/1:review` for the path to see what I'm trying to get. You'll also want to check out [Google's API style guide page on custom verbs](https://cloud.google.com/apis/design/custom_methods).

Basically, in order to have a colon in the url and because the url path matcher package uses colons as a special character for path variables, we have to escape the colon and in order to escape the colon we have to escape the escape character to keep the escape character in the string.

Admittedly, yes, this is very confusing and should probably be documented somewhere, but I haven't thought too much about it.

kevinyu328: Ahh I see. Understood, thanks for clarifying!

kevinyu328 (approved):
Looks good!

PR #134, opened 2021-04-11
Added central frontend routes file

### LINKED ISSUE #92: General - Centralize Application Front-End Routes in Shared Folder
To facilitate the better loose coupling between various different parts of the application when it comes to the layout of all the different application routes for the front-end, we would like to define all of them in a central routes file.

jamescd18:
Closes #92. Created centralized file in shared folder for defining front-end routes.


kevinyu328 (approved):
Nice

PR #143, opened 2021-04-16
Lifted the router up to index.tsx for authentication

jamescd18:
Lifted the browser router up into the `index.tsx` file in order to allow the login page to be displayed without the top navigation bar. Potentially will want to rethink the top-most layers of `index.tsx > App.tsx > pages...` given that the items necessary at the `index.tsx` level will increase over time. Things like react context and testing the top-most level will potentially get a little tricky, and it seems like not the best practice to have a lot of logic in `index.tsx`.

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -3,18 +3,28 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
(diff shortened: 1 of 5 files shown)


kevinyu328 (approved):
Nice! True, will probably have to create some more top-level pages/components

PR #144, opened 2021-04-16
Remove body tag and restyle 404 page

jamescd18:
Removed the body tag from the 404 page not found component. Also went ahead and restyled the component, changed to using class names for CSS styling, and added top padding.

Screen Shot 2021-04-16 at 9 41 10 AM


kevinyu328 (approved):
Looks good!

PR #145, opened 2021-04-16
Fix manufacturing spelling mistake

jamescd18:
Fixed the little spelling mistake
Screen Shot 2021-04-16 at 9 50 12 AM


kevinyu328 (approved):
Good catch

Issue #124, opened 2021-03-28
Auth - Lift Router into index.tsx to Route Login


jamescd18:
Currently, the react router is located within the app.tsx file. This means that all routes will always contain the nav links present in the app.tsx. We want to display the login component without the nav links. To do this, let's lift the the react router to index.tsx. This will mean that there will be a switch within index.tsx which contains only two routes: one for login, and one that matches anything else and displays the app.tsx component. App.tsx will maintain the switch, nav links, and sub-routes.

PR #149, opened 2021-04-17
Implemented Basic React Context for User Auth

### LINKED ISSUE #125: Auth - Configure React Context for User Authentication
Basically, we're going to user React context to help us with authentication on the front-end. If not logged in, send to login page, otherwise, go to home page. Check out [this article for a quick tutorial on react context](https://dev.to/curtiscodes/learn-to-usecontext-with-hooks-in-3-minutes-4c4g).

jamescd18:
Implemented a basic version of the react context configuration required to support an authenticated user on the front-end. Decided that the react router would not fit in the `index.tsx` file after all (despite #124) and instead the `App.tsx` component would need to be expanded and split apart. The solution was to create three components: `AppMain`, `AppPublic`, and `AppAuthenticated`. The `AppMain` component handles the react context and checks for a logged in user to decide whether to display the public or authenticated pages of the app. The `AppPublic` component just displays the login page for all routes, but opens the app to future expansion if additional public pages are desired. The `AppAuthenticated` component displays the normal application view with the top bar and pages below. 

The user flow is as follows: go to the app > land on the login page > type in name and hit enter > go to authenticated part of app. If the user reloads the page, they are effectively logged out. Clicking on the user menu in the navigation bar in the authenticated part of the app will display the user's name to indicate they are logged in. The log out button logs the user out and sends them to the home page, whereby they are redirected to the login page.

Closes #125 

Login page
Screen Shot 2021-04-16 at 10 23 11 PM

Logged in
Screen Shot 2021-04-16 at 10 23 34 PM


kevinyu328 (approved):
Nice, looks good!! I like the little Easter egg put in for Rachel

PR #152, opened 2021-04-17
Add API_URL to utils and updated projects endpoint

### LINKED ISSUE #151: General - Update API Route Matching
It looks like with the AWS Lambda updates, Netlify also changed the `path` variable that we were previously matching API routes against. Now, the `route` always seems to include the beginning portion related to the API that is `/.netlify/functions`. You can see this in action on #38 with the latest screenshot.

jamescd18:
Add API_URL to utils and updated projects endpoint. Closes #151


kevinyu328 (approved):
Nice

PR #166, opened 2021-04-24
Establish Database Seeding Process

### LINKED ISSUE #114: General - Create a Database Seed Process
During development, we will need a quick and easy way to seed the database with data in order to more easily interact with the application in meaningful ways (it's a data-driven data display and data manipulation application). We can fairly easily accomplish this by executing a "reset database" SQL script file (`reset_db.sql`) via the `psql` Postgres CLI tool. Then we can also execute a TypeScript script that uses Prisma and the data types to instantiate a nice set of test data.

Legacy, not valid:
- [x] Create a seed folder inside of the src/backend folder to house everything related to seeding the database
- [x] Create a seed SQL script file
   - [x] Add the standard repository license comment
   - [x] Add drop database, create database statements
- [x] Create a TypeScript file to define the actual seeding process

jamescd18:
Went through the process of digging into everything Prisma and understanding how the built-in seeding functionality works. Basically, we should just be able to write everything in the seed file and then run the seed command and Prisma will drop, recreate, and seed the database. Closes #114.


eshwaribhide (approved):
Looks fantastic and awesome and great and wonderful and amazing

PR #171, opened 2021-04-26
Add routeMatcher to users endpoint

### LINKED ISSUE #148: Auth - Expand Users Endpoint with RouteMatcher Pattern
Now that we have established the route matcher pattern for API routing via the projects endpoint, we need to apply this to the users endpoint. This will enable us to expand the endpoint beyond just returning a list of all the users.

jamescd18:
Added routeMatcher to users endpoint. Added single user route to API routes. Closes #148

### DIFF
--- a/src/backend/functions/users.ts
+++ b/src/backend/functions/users.ts
@@ -4,20 +4,59 @@
  */
 
 import { Context } from 'aws-lambda';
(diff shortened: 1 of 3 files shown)


eshwaribhide (approved):
Looks fantastic and awesome and great and wonderful and amazing

PR #175, opened 2021-04-30
Major changes to Prisma schema and application types

jamescd18:
Major changes to update the Prisma schema and overhaul parts of the application types


jamescd18:
Also I deleted and recreated the migrations. Might do it again later but we'll see. Migrations are really only useful after we've got production data loaded in or something since they enforce iterative changes. Experiment and learn as we go.

Issue #159, opened 2021-04-21
Projects - Create the Work Packages Rules Component


jamescd18:
Create the view-only component for displaying work package rules references. This should cleanly and readably display each element nicely spaced within a card component. Refer to the design sketch v4 snippet below for inspiration and context.

----

Here's a quick outline of steps for this issue. Check them off and / or suggest edits to them as you make progress on the issue. Existing dummy components may be useful to consult for guidance ([see here](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/src/components/change-request-details/change-request-details.tsx), and [here](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/91)).

- [ ] Create the component folder in `/src/components/`
- [ ] Create the test file, component file, and ccs module in the component folder
- [ ] Add basic TS functional component boilerplate to component file
- [ ] Add basic test suite `describe()` to test file
- [ ] Add the component to the `src/app/app-authenticated` component (so you can see the component when starting the app in development)
- [ ] Configure component props to take in the proper work package rules string array
- [ ] Display all desired component props in the tsx
- [ ] Add DOM element division to begin arranging / styling data elements
- [ ] Add styling to CSS module file
- [ ] Attach CSS styling to proper sections / DOM elements in tsx
- [ ] Get the component to an aesthetically pleasing state that resembles the design sketch snippet below
- [ ] Open a PR with a screenshot of the component to get review and feedback
- [ ] Iterate and make changes based on feedback
- [ ] Once approved for merging, attach a screenshot of the finished component to this issue

---
Design Sketch v4 Snippet: See the card titled "Rules"
<img width="809" alt="Screen Shot 2021-04-21 at 9 43 33 AM" src="https://user-images.githubusercontent.com/51571627/115590627-0a79bd80-a286-11eb-9104-f1ad55bffa85.png">

djung335:
![image](https://user-images.githubusercontent.com/44933949/117069713-b0471680-ace1-11eb-909b-fec00d708859.png)

PR #182, opened 2021-05-03
#159 - Created the work package rules component

djung335:


jamescd18 (changes requested):
There's 3 unused imports that need to be removed, but otherwise this looks great!

jamescd18 (approved):

PR #186, opened 2021-05-04
Create data for seeding database

### LINKED ISSUE #185: General - Create Example Data Specific For Seeding
The database requires different fields, different types, and different arrangements of data which is specific to seeding the database. Let's create a new folder `src/backend/prisma/seed-data` to house data that is specifically built for seeding the database. This example data will have to span the entire range of tables in the database.

jamescd18:
Created several files of data objects specific for seeding the database. Utilized arrays and flexible seed object creation in order to allow easier expansion in the future for different values, additional database entries, and more. Closes #185


kevinyu328 (approved):
Nice nice

PR #223, opened 2021-05-14
219- home page component

djung335:
I've gotten everything except for the testing for the home page down. I want to display the user's name, but I can't seem to pass that in anywhere.


djung335:
image

jamescd18 (changes requested):
Looks pretty good overall honestly. You'll want to change the "Welcome" check expect to do `/Welcome/i` instead of `'Welcome'` and see if that will make it pass

jamescd18 (approved):
Awesome job!

PR #224, opened 2021-05-17
Separate the projects-table container into a container

### LINKED ISSUE #137: Projects - Separate the projects-table container into a container + a view-only component
In order to reduce the complexity of each individual react component, we want to try and separate different functionality between view-only components and container components. Currently, the projects-table component was built before realizing that separation would be beneficial for this component. The goal of this issue is to separate the view functionality from the data fetching and transformation functionality.

Basically, we want to put the table configuration details in a view-only component that accepts the array of display projects as a component prop. Then the data fetching and transformation logic will stay in the projects-table container component.

TheFishToldMeSo:
closes #137 
Separate the projects-table container into a container


jamescd18 (changes requested):
Same issues with package-lock.json

TheFishToldMeSo (commented):
Deleted package-lock.json from commit

jamescd18:
I don't think you want to be deleted the package-lock.json......

We just don't want to change it.

PR #225, opened 2021-05-17
Projects - Update to Use React-Router Link

### LINKED ISSUE #158: Projects - Update to Use React-Router Link
Currently, some locations may be using the link pipe when the desired behavior is actually a react-router link navigation. Replace the below example that use the link pipe and instead have them utilize the `Link` component from react-router. See [here](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/78c2714e109ab37814f3facaa04b7b0c3d163a6f/src/components/nav-page-links/nav-page-links.tsx#L16) for an example usage of the `Link` component.

- [The work packages summary](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/78c2714e109ab37814f3facaa04b7b0c3d163a6f/src/components/work-package-summary/work-package-summary.tsx#L36)

TheFishToldMeSo:
closed #158 
Update to Use React-Router Link


jamescd18 (changes requested):
Oh dear. Make sure you don't commit changes to the package-lock.json unless you change the package.json. I might suggest creating a new branch and redoing the changes instead of trying to reverse the changes to the package-lock.json files, but it's up to you.

PR #227, opened 2021-05-24
Enhance plain text login

### LINKED ISSUE #221: Auth - Create Stored Plain Text Login
Very initial MVP for user authentication will start with the following:
- Storing the user `emailId` in local storage
- Storing the user `emailId` in browser memory for react context
- Storing the page the user was on before redirecting to the login page
- NO connection to the API

See #121 for the workflow diagram sketches.

jamescd18:
- added lots of tests
- mock `localStorage` in tests
- split app section so context is separate from login logic
- enabled storage of userId in `localStorage`
- enable storage of page before login for redirect after login
- closes #221

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,9 @@
       "!<rootDir>/src/utils/node_modules/",
       "!**/lib/**",
       "!src/**/*.d.ts"
+    ],
+    "setupFilesAfterEnv": [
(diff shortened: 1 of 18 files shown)


kevinyu328 (approved):
Nicee

PR #234, opened 2021-06-03
#215 add new custom hooks

dylan-pham:
I've updated the code implementation but still have to update the tests.

### DIFF
--- a/src/containers/projects-table/projects-table.tsx
+++ b/src/containers/projects-table/projects-table.tsx
@@ -14,8 +14,8 @@ const ProjectsTable: React.FC = () => {
 
   /**
    * Transform given list of projects into a list of display projects.
-   * @param
-   * @returns
+   * @param projects list of projects to transform
+   * @returns list of display projects
    */
   const updateData = (projects: Project[]) => {
     return projects.map((prj: Project) => {
--- a/src/containers/projects-table/projects-table.tsx
+++ b/src/containers/projects-table/projects-table.tsx
@@ -3,54 +3,48 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
-import { useEffect, useState } from 'react';
-import { AxiosResponse } from 'axios';
 import { Project, WorkPackage } from 'utils';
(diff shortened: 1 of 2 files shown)


src/containers/projects-table/projects-table.tsx
```diff
@@ -3,54 +3,48 @@
...
-      })
-    );
+  const { isLoading, errorMessage, responseData } = useAllProjects();
+
+  /**
+   * Transform given list of projects into a list of display projects.
+   * @param projects list of projects to transform
+   * @returns list of display projects
+   */
+  const updateData = (projects: Project[]) => {
```
jamescd18, line 20: I might suggest a better name for this function like `projectsToDisplayProjects()` or `transformToDisplayProjects()`

src/containers/projects-table/projects-table.tsx
```diff
@@ -3,54 +3,48 @@
...
+      <>
+        <h3>Oops, sorry!</h3>
+        <h5>There was an error loading the page.</h5>
+        <p>{errorMessage ? errorMessage : 'The data did not load properly.'}</p>
+      </>
+    );
+  }
 
-  return <PrjsTable allProjects={allProjects} />;
+  return <PrjsTable allProjects={updateData(responseData)!} />;
```
jamescd18, line 47: Any idea why the ! is on the right side of the closing parentheses? I'm not quite familiar enough / maybe not remembering. Does it not work putting the ! to the left of the closing parentheses?

jamescd18 (changes requested):
Looking good! These behavior changes look to be correct. I left a couple suggestions / curiosities.

src/containers/projects-table/projects-table.tsx
dylan-pham: I believe including the ! on the right side is telling the compiler that the return value of the function call is non-null whereas including it on the left side just means that responseData is non-null. I honestly hadn't seen this syntax before so I had to do some googling.

PR #237, opened 2021-06-04
#220 login endpoint

### LINKED ISSUE #220: Auth - Create User Login Endpoint
Need to create an endpoint and front-end service for logging users in. Should take in an emailId string, check if user exists, log user in, return user object, return error if the user does not exist.

### LINKED ISSUE #231: Tests - Migrate API Tests to Use test-support/test-data
The API tests are still using test artifacts from the `utils` package, so let's reduce their dependency on the `utils` package and migrate over to using the data provided in the `src/test-support/test-data` folder.

### LINKED ISSUE #232: General - Clean Up the Utils Package
- remove `example-shared.ts`
- delete api-utils test data from `dummy-data.ts`

jamescd18:
- Created login endpoint in the API
- abstracted user object field assertions
- deleted example shared from utils package
- moved API testing utilities from utils package to `test-support` folder
- modified `mockEvent()` to accept request bodies

- closes #232
- closes #220 
- closes #231


kevinyu328 (changes requested):
Just a couple of unused imports in dummy-data.ts, but other than that, everything looks good to me

kevinyu328 (approved):
Nice

Issue #200, opened 2021-05-10
Change Requests - Switch Change Requests Table to Front-End Service Hook


jamescd18:
**Background:**
In order to increase flexibility, simplicity, and separation of behavior in the connection between our front-end and back-end, we need our container components to adopt the usage of the new custom React hooks created in #116 and #210. To do this, we need container components to implement the proper hooks instead of executing `apiFetch` requests themselves.

---

**Behavior:**
Container components will need to have their current fetching code replaced with usage of the custom hooks as is seen in the `src/pages/change-request-details/change-request-details.tsx` file (currently seen in #210). Following the design pattern of the new React hooks, container components will need to handle the following major states:
- Display loading text when the request is loading
- Display error text when the request returns an error message
- Display the correct component / desired behavior when the data is successfully returned

---

**Steps:**
- [ ] Change the code implementation
  - [ ] Remove the old fetching code
  - [ ] Add the single line for the proper custom hook
  - [ ] Add the logic for the loading state
  - [ ] Add the logic for the error state
  - [ ] Add the logic for the normal successful return state
- [ ] Update the tests
  - [ ] Utilize the mocking pattern found on lines 14 to 23 of `src/pages/change-request-details/change-request-details.test.tsx` (currently seen in #210)
  - [ ] Remove mocking logic implemented using the `msw` package
  - [ ] Properly mock the hook in every test case

---

Linked to: #200, #215, #216

PR #246, opened 2021-06-09
#200 yoar

YoAr:
The changes container updates for Change request table container to new custom hook pattern along with updated test cases . 
The test cases have some issues with the example data where a nudge from anyone would be helpful for me.
https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/200

### DIFF
--- a/src/components/change-requests-table/change-requests-table.tsx
+++ b/src/components/change-requests-table/change-requests-table.tsx
@@ -10,19 +10,9 @@ import BootstrapTable, {
   SortOrder
 } from 'react-bootstrap-table-next';
 import styles from './change-requests-table.module.css';
-
-export interface DisplayChangeRequest {
-  id: number;
-  submitterName: string;
-  wbsNum: string;
-  type: string;
-  dateReviewed: string;
-  accepted: string;
-  dateImplemented: string;
-}
-
+import { ChangeRequest } from 'utils/src';
 interface ChangeRequestsTableProps {
-  changeRequests: DisplayChangeRequest[];
+  changeRequests: ChangeRequest[];
 }
 
 /**
--- a/src/containers/change-requests-table/change-requests-table.test.tsx
+++ b/src/containers/change-requests-table/change-requests-table.test.tsx
@@ -10,7 +10,7 @@ import { ApiHookReturn } from '../../services/api-request';
 import { ChangeRequest,exampleAllChangeRequests } from 'utils';
 import { routes } from '../../shared/routes';
 import { renderWithRouter } from '../../test-support/test-utils'
-import { wbsPipe } from '../../shared/pipes';
+import { fullNamePipe, wbsPipe } from '../../shared/pipes';
 
 
 jest.mock('../../services/change-requests');
@@ -55,7 +55,7 @@ describe('change requests table container', () => {
     renderComponent();
     expect(screen.queryByText('Loading...')).not.toBeInTheDocument();
     expect(screen.getByText(exampleAllChangeRequests[0].id, { exact: false })).toBeInTheDocument();
-    expect(screen.getByText(exampleAllChangeRequests[0].submitterName!)).toBeInTheDocument();
+    expect(screen.getByText(fullNamePipe(exampleAllChangeRequests[0].submitter)!)).toBeInTheDocument();
     expect(screen.getByText(wbsPipe(exampleAllChangeRequests[0].wbsNum))).toBeInTheDocument();
     expect(screen.getByText(exampleAllChangeRequests[0].type)).toBeInTheDocument();
     expect(screen.getByText(exampleAllChangeRequests[0].dateReviewed!.toLocaleDateString())).toBeInTheDocument();
--- a/src/containers/change-requests-table/change-requests-table.test.tsx
+++ b/src/containers/change-requests-table/change-requests-table.test.tsx
@@ -3,32 +3,45 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
-import { rest } from 'msw';
-import { setupServer } from 'msw/node';
-import { render, screen } from '@testing-library/react';
-import { exampleAllChangeRequests } from '../../test-support/test-data/change-requests.stub';
+import { screen } from '@testing-library/react';
 import ChangeRequestsTable from './change-requests-table';
+import { useAllChangeRequests } from '../../services/change-requests';
+import { ApiHookReturn } from '../../services/api-request';
+import { ChangeRequest,exampleAllChangeRequests } from 'utils';
+import { routes } from '../../shared/routes';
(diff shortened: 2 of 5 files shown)


src/components/change-requests-table/change-requests-table.tsx
```diff
@@ -10,19 +10,9 @@ import BootstrapTable, {
...
-  wbsNum: string;
-  type: string;
-  dateReviewed: string;
-  accepted: string;
-  dateImplemented: string;
-}
-
+import { ChangeRequest } from 'utils/src';
 interface ChangeRequestsTableProps {
-  changeRequests: DisplayChangeRequest[];
```
jamescd18, line 25: Why remove the `DisplayChangeRequest` interface? This interface defines what data this component needs in order to properly display data in the bootstrap-react table.

src/containers/change-requests-table/change-requests-table.test.tsx
```diff
@@ -3,32 +3,45 @@
...
 import ChangeRequestsTable from './change-requests-table';
+import { useAllChangeRequests } from '../../services/change-requests';
+import { ApiHookReturn } from '../../services/api-request';
+import { ChangeRequest,exampleAllChangeRequests } from 'utils';
+import { routes } from '../../shared/routes';
+import { renderWithRouter } from '../../test-support/test-utils'
+import { fullNamePipe, wbsPipe } from '../../shared/pipes';
 
-const endpointURL: string = '/.netlify/functions/change-requests';
+
```
jamescd18, line 15: This is great! Glad to see you using the jest.mock pattern here for testing.

src/containers/change-requests-table/change-requests-table.test.tsx
```diff
@@ -3,32 +3,45 @@
...
+const mockHook = (isLoading: boolean, errorMessage: string, responseData?: ChangeRequest[]) => {
+  mockedUseAllChangeRequest.mockReturnValue({ isLoading, errorMessage, responseData });
+};
 
 // Sets up the component under test with the desired values and renders it.
 const renderComponent: () => void = () => {
-  render(<ChangeRequestsTable />);
+  renderWithRouter(<ChangeRequestsTable />, {
+    path: routes.CHANGE_REQUESTS,
+    route: `${routes.CHANGE_REQUESTS}`
```
jamescd18, line 31: You should be able to remove the template string surrounding `routes.CHANGE_REQUESTS` since the template string isn't adding anything.

src/containers/change-requests-table/change-requests-table.tsx
```diff
@@ -3,58 +3,27 @@
...
+  if (isLoading) {
+    return <p>Loading...</p>;
+  }
+  if (errorMessage !== '' || responseData === undefined) {
+    return (
+      <>
+        <h3>Oops, sorry!</h3>
+        <h5>There was an error loading the page.</h5>
+        <p>{errorMessage ? errorMessage : 'The data did not load properly.'}</p>
+      </>
```
jamescd18, line 23: Nice usage of the new custom hook design pattern!

src/containers/change-requests-table/change-requests-table.tsx
```diff
@@ -3,58 +3,27 @@
...
-
-    // cleanup function indicates component has been unmounted
-    return () => {
-      mounted = false;
-    };
-  }, []);
-
-  return <CRTable changeRequests={allChangeRequests} />;
+  }
+  return <CRTable changeRequests={responseData!} />;
```
jamescd18, line 26: This component should take in an array of `DisplayChangeRequest`s last I checked. This means that you will need to transform the response data array into an array of `DisplayChangeRequest` in a very similar way to the code you deleted. See comment above.

src/containers/change-requests-table/change-requests-table.tsx
```diff
@@ -3,58 +3,27 @@
...
-          submitterName: fullNamePipe(cr.submitter),
-          wbsNum: wbsPipe(cr.wbsNum),
-          type: cr.type,
-          dateReviewed: cr.dateReviewed ? new Date(cr.dateReviewed).toLocaleDateString() : '',
-          accepted: cr.accepted ? booleanPipe(cr.accepted) : '',
-          dateImplemented: cr.dateImplemented
-            ? new Date(cr.dateImplemented).toLocaleDateString()
-            : ''
-        };
-      })
```
jamescd18, line 33: I believe this is the code that you'll need to use to transform a `ChangeRequest` array into a `DisplayChangeRequest` array.

src/services/change-requests.ts
```diff
@@ -17,13 +18,15 @@ import { useApiRequest } from './api-request';
...
-      ? new Date(changeRequest.dateReviewed)
+      ? new Date(changeRequest.dateReviewed).toLocaleDateString()
       : changeRequest.dateReviewed,
     dateImplemented: changeRequest.dateImplemented
-      ? new Date(changeRequest.dateImplemented)
-      : changeRequest.dateImplemented
+      ? new Date(changeRequest.dateImplemented).toLocaleDateString()
+      : changeRequest.dateImplemented,
+      submitterName: fullNamePipe(changeRequest.submitter),
+      wbsNum: wbsPipe(changeRequest.wbsNum)
```
jamescd18, line 29: I don't believe this code should be changed. The purpose of this function is only to ensure that field which are supposed to be date objects are properly transformed to date objects. (There's an issue where they seem to stay as date strings, which is undesirable.)

src/services/change-requests.ts
```diff
@@ -32,6 +35,10 @@ export const changeRequestTransformer = (changeRequest: ChangeRequest) => {
   return output;
 };
 
+export const allChangeRequestTransformer = (changeRequests: ChangeRequest[]) => {
```
jamescd18, line 38: Why does this function need to be exported? I might have missed where it is used, but I'm pretty sure that keeping it inline like it was before is better.

jamescd18 (changes requested):
This is a great start! There's a few things that are confusing me about the changes you've made, and I've added them as inline comments. Also, there's a few pointers I've outlined below.

- Are you using VSCode? If so, have you got the configurations set up properly? Specifically, is Prettier set up according to [the instructions](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/docs/Onboarding.md#ide-vscode)?
- Try to follow the commit syntax specified in the [contributor guide](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/docs/ContributorGuide.md#creating-a-commit)
- It seems to me like the test failures are due to modifications to the data structures, something that I don't believe is necessary for this ticket. I've added more details in inline comments.

Let me know if any of my comments, suggestions, or feedback don't make sense. I'm always happy to sit down and chat to help ensure we're both understanding what's going on!

PR #251, opened 2021-06-14
#205 - Restructure Files Into Feature Tree

### LINKED ISSUE #205: General - Restructure Files into a Feature-Based Tree
**Background:**
When we originally set up the project repository, we decided to give it a try with components separated by their behavior type of pages, containers, and presentational view-only components. This has turned out to be a ton more confusing than initially anticipated for the React front-end. As such, we would like to change to a feature-based file structure that is a lot more logical and nested.

---

**Steps:**
- [ ] Rearrange all the files

kevinyu328:
Restructured files into a feature-based tree. Was unsure if the 'nav-top-bar' folder should be within the 'shared' folder or have its own separate folder. Closes #205.


jamescd18 (commented):
It looks like tests are failing because of file path changes. The `jest.Mock()` method invocations rely on file paths so they will also need to be updated.

jamescd18 (changes requested):
This is great! Just a couple things I think we should make sure to do:
- Can you pull the app folder into the components folder? I don't think anything inside the app folder should be restructured, but the app folder also contains components. You'll likely need to make sure to edit the index.tsx file, and the file paths of the imports in the app folder
- I think the top nav shouldn't be in the shared folder, and it's okay (if it works) to have the shared folder be empty
- Can you also open a separate ticket to account for the renaming we should do to eliminate where we currently have components with the same names?

jamescd18 (approved):

PR #275, opened 2021-07-12
I267 leroy - updating settings page

leoRysing:
I did some fiddling with incorporating the PageBlock and PageTitle Components into the settings page. I think the use of the components works, but I definitely think some review here would be good, as well as the usage of the created components on the page.


src/components/settings/settings.tsx
```diff
@@ -7,12 +7,22 @@ import { useContext } from 'react';
 import { UserContext } from '../app/app-context/app-context';
 import './settings.module.css';
 
+import PageTitle from '../shared/page-title/page-title';
```
jamescd18, line 10: Place new imports right about the CSS file import

src/components/settings/settings.tsx
```diff
@@ -7,12 +7,22 @@ import { useContext } from 'react';
...
+import PageBlock from '../shared/page-block/page-block';
+
+// †ødø; add the two new components, Page Title and Page Block, from shared
```
jamescd18, line 13: Update to do or remove it

src/components/settings/settings.tsx
```diff
@@ -7,12 +7,22 @@ import { useContext } from 'react';
...
+
 const Settings: React.FC = () => {
   const user = useContext(UserContext);
+  const title = <PageTitle title="This is the Settings Page" />
```
jamescd18, line 17: Place the value of title inline with the return to simplify

src/components/settings/settings.tsx
```diff
@@ -7,12 +7,22 @@ import { useContext } from 'react';
...
+  // const body = PageBlock("User Settings", "", "User: {user}");
```
jamescd18, line 18: Remove extraneous comments

src/components/settings/settings.tsx
```diff
@@ -7,12 +7,22 @@ import { useContext } from 'react';
...
+  // render(<PageBlock title={'test'} headerRight={<>hi</>} body={<>hello</>} />);
+  const pageBlockBody = <>User: {user}</>;
```
jamescd18, line 20: You can inline this too

src/components/settings/settings.tsx
```diff
@@ -7,12 +7,22 @@ import { useContext } from 'react';
...
+  const body = <PageBlock title="User Settings" headerRight={<></>} body={pageBlockBody} />
```
jamescd18, line 21: Inline this too

jamescd18 (changes requested):
Great work! It looks to be functional, there's just some code quality and cleanup items too

src/components/settings/settings.tsx
leoRysing: Quick question, does this just mean make the imports a single block?

jamescd18: Yes, all the imports should be together in one block at the top of the file. The css file import is almost always the last import.

jamescd18:
Any updates, @leoRysing?

jamescd18 (dismissed):

jamescd18 (approved):
Awesome! Glad to see it!

PR #280, opened 2021-08-14
Massive overhaul for authentication implementation

### LINKED ISSUE #250: Auth - Separate Login Component
In order to increase flexibility for future expansion, the login component needs to be separated into two components: one for the logic of logging in, and another for the view-only elements that are displayed in the UI. This will drastically improve our ability to add features and keep the code clean.

### LINKED ISSUE #276: General - Convert Back-End Methods to Async
Adjust the back-end so that asynchronous calls are awaited at the last moment possible. This will increase flexibility when it comes to writing asynchronous code in the back-end.

### LINKED ISSUE #278: Auth - Connect User Login with Google Auth API
To avoid the effort of having to implement a lot of different features related to security, password reset, two-factor authentication, etc., we would like to use the Google authentication APIs to allow users to log in with their Google credentials.

### LINKED ISSUE #279: Auth - Switch to Auth Pattern Provided by React Router
React Router provides a handy example for implementing authentication in your front-end application. We're going to do our best to model this same pattern for the application, with the hopes of landing with a design pattern which is stable enough for production.

- https://reactrouter.com/web/example/auth-workflow
- https://usehooks.com/useAuth/

jamescd18:
Changes
- Major refactoring of front-end user context design pattern to match example provided by react router
- Large adjustments to login endpoint to accommodate google auth
- All backend endpoints can now handle promises or direct return
- Refactoring login pages to better handle google auth and separate dev login

Tickets
- Closes #276 
- Closes #278 
- Closes #250 
- Closes #279


kevinyu328 (changes requested):
Looks good! Just two small linting errors, but that's it.

PR #285, opened 2021-08-14
#271 - changing mouse hovering effect for project table into nice finger

### LINKED ISSUE #271: General - Make Mouse Hover Be Finger Click
Currently, the rows in the tables on the Projects and Change Requests tables are all clickable, but when a user hovers their mouse over a row, the mouse icon stays the same. 

In order to improve the user experience, we should make sure that when users hover over a row in these tables, the mouse pointer icon changes to the finger one that indicates the row is clickable.

VerySeriousHireableAccount:
closes #271 changing mouse hovering effect for project table into nice finger


jamescd18 (approved):
Amazing! Looks great!

PR #315, opened 2021-08-24
#212 project connect db

jamescd18:
Experimenting and figuring out what the proper configuration and design patterns would be for establishing connections from the backend to the database via Prisma. This is most challenging in the department of figuring out how to deal with Prisma's type-safe responses. Essentially, determining what types from Prisma are okay to use where in the application, and how to translate from Prisma types into application types depending on the scenario. Merging this will begin to cause production to break given that we have not configured a production or staging database for Netlify to connect to.


jamescd18:
As expected, the utils package does not actually have access to everything provided by the Prisma packages. More effort will have to go into restructuring the linkage between application types and database types given that database types are effectively only available in single endpoint files or in the front-end.

PR #323, opened 2021-08-28
#214 connect users db

### LINKED ISSUE #214: Auth - Connect the Users API Endpoint to the Database
**Background:**
In order to move towards supporting dynamic data, we need the API to pull data from the database instead of only serving up the hard-coded data provided by the dummy data from the utils package. To do this, we need to the API to query the database via Prisma.

---

**Behavior:**
At this point in time, we do not have a staging or product database set up yet. As such, the database is only available in local development. For initially connecting the API endpoint to the database, we will have the API methods check if the incoming request came from the application running locally or not. If the request came from localhost, query from the database. Otherwise, supply the hardcoded example dummy data as we are doing right now.

---

**Steps:**
- [ ] Modify the API methods to take in event as the second argument (params is the first one)
- [ ] Add a conditional check on `event.headers.host` to see if it's equal to `localhost:3000`
- [ ] Query the database via Prisma

---

Linked to: #198, #212, #213, #214

jamescd18:
Connecting the users endpoint to the database, hopefully turns out better than my first pass at connecting the projects to the database. Closes #214.


jamescd18:
Switch to feature branch `database-setup` like in #324 and #325

PR #326, opened 2021-08-28
Database setup

jamescd18:
Feature branch for setting up the database connection to the backend properly because it is messy. Related to #212, #214.


jamescd18:
The application tests now pass (locally) and the application will run. There are still glaring issues that need to be resolved when it comes to supplying the right fields in API response objects and having the type interfaces agree with each other.

Particularly, the pattern where we have a project summary interface with a narrower set of fields is what we need to replicate for other parts of the application. Either that, or we have to expand the database queries to provide more and more nested data. I'm more in favor of developing more specific interfaces, because that decouples the data type for one page in the application from the data type for another page, thereby opening up to future flexibility where you can update the prisma query, data type interface, transformer, and front-end in one traceable stack without having to bother too much with accidentally breaking another page in the application.

Particularly, a change request summary interface is needed, and a work package summary interface is needed. The lack of these breaks the change requests table page and the individual project page.

src/utils/src/backend-supports/projects-get-all.ts
```diff
@@ -0,0 +1,29 @@
+/*
```
acd124, line 1: This file and its functions are not referenced anywhere. This may indicate that the backend functions are doing things manually that should be abstracted here.

acd124 (changes requested):
The calculation of project duration isn't as specified in the backend supports tests. Also project leads are not currently displaying properly or getting sent to the frontend properly. This will likely require further research to fix allowing multiple project leads.

src/utils/src/backend-supports/projects-get-all.ts
jamescd18: Let's leave them as examples and file ticket(s) for improvements to the backend.

jamescd18: See #411

src/components/projects/projects-table/projects-table.tsx
```diff
@@ -126,11 +112,9 @@ const ProjectsTable: React.FC = () => {
     const leads: User[] = [];
     const seenList: number[] = [];
     for (const project of projects) {
-      for (const user of project.projectLead) {
-        if (!seenList.includes(user.id)) {
-          seenList.push(user.id);
-          leads.push(user);
-        }
+      if (!seenList.includes(project.projectLead!.userId)) {
```
acd124, line 115: This fails when the project lead is missing. Optional chaining is likely needed, but not null assertion can be used at the end to trick the type checker into agreeing that things make sense.

acd124 (changes requested):
Tests in `src/utils/tests/api-utils.test.ts` now seem to be showing a type error since the return type for `routeMatcher` can now be a Promise. This does not seem to break anything or fail any tests, but should be addressed. Other than that, just need to fix the optional project lead.

acd124 (approved):

Issue #96, opened 2021-03-06
Projects - Fix Projects Table Sorting


jamescd18:
A few leftover items need to be cleaned up with the Projects Table container component.

- [x] Ensure column headers are aligned center via `headerAlign: 'center'` in the column configs
- [x] Ensure wbsNums are sorted properly first by area, then project, then work package
- [x] Ensure duration is properly sorted

PR #334, opened 2021-09-20
#96 Projects Table Sorting

shanejpark:
We've resolved all of the issues that were presented in #96. The first one, centering the tables, was pretty simple to fix, just adding headerAlign = 'center' to each column and changing the align from 'left' to 'center'. To fix the version sorting, we had to override the default sorting algorithm and implement our own with sortFunc. The only one I'm not sure about is the Duration sort, since it seems like it's already been resolved.

### DIFF
--- a/src/components/projects/projects-table/projects-table/projects-table.test.tsx
+++ b/src/components/projects/projects-table/projects-table/projects-table.test.tsx
@@ -138,5 +138,3 @@ describe('projects table component', () => {
     expect(wbsNumsAsc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder.reverse());
   });
 });
-
-
--- a/src/components/projects/projects-table/projects-table/projects-table.test.tsx
+++ b/src/components/projects/projects-table/projects-table/projects-table.test.tsx
@@ -116,4 +116,27 @@ describe('projects table component', () => {
     const wbsNumsAsc: HTMLElement[] = await screen.findAllByText(wbsRegex);
     expect(wbsNumsAsc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder.reverse());
   });
+
+  it('handles sorting and reverse sorting the table by wbsNum', async () => {
+    renderComponent(exampleAllProjects);
+
+    const column: string = 'WBS #';
+    const expectedWbsOrder: string[] = [
+      exampleProject5,
+      exampleProject4,
+      exampleProject3,
+      exampleProject2,
+      exampleProject1
+    ].map((prj: Project) => wbsPipe(prj.wbsNum));
+
+    fireEvent.click(screen.getByText(column));
+    const wbsNumsDesc: HTMLElement[] = await screen.findAllByText(wbsRegex);
+    expect(wbsNumsDesc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder);
+
+    fireEvent.click(screen.getByText(column));
+    const wbsNumsAsc: HTMLElement[] = await screen.findAllByText(wbsRegex);
+    expect(wbsNumsAsc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder.reverse());
+  });
 });
+
+
--- a/src/components/projects/projects-table/projects-table/projects-table.test.tsx
+++ b/src/components/projects/projects-table/projects-table/projects-table.test.tsx
@@ -117,7 +117,7 @@ describe('projects table component', () => {
     expect(wbsNumsAsc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder.reverse());
   });
 
-  it.skip('handles sorting and reverse sorting the table by wbsNum', async () => {
+  it('handles sorting and reverse sorting the table by wbsNum', async () => {
     renderComponent(exampleAllProjects);
 
     const column: string = 'WBS #';
--- a/src/components/projects/projects-table/projects-table/projects-table.test.tsx
+++ b/src/components/projects/projects-table/projects-table/projects-table.test.tsx
@@ -117,7 +117,7 @@ describe('projects table component', () => {
     expect(wbsNumsAsc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder.reverse());
   });
 
-  it('handles sorting and reverse sorting the table by wbsNum', async () => {
+  it.skip('handles sorting and reverse sorting the table by wbsNum', async () => {
     renderComponent(exampleAllProjects);
 
     const column: string = 'WBS #';
--- a/src/components/projects/projects-table/projects-table/projects-table.tsx
+++ b/src/components/projects/projects-table/projects-table/projects-table.tsx
@@ -3,7 +3,7 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
-import {useHistory} from 'react-router-dom';
+import { useHistory } from 'react-router-dom';
 import BootstrapTable, {
   ColumnDescription,
   RowEventHandlerProps,
@@ -11,7 +11,7 @@ import BootstrapTable, {
 } from 'react-bootstrap-table-next';
 import PageTitle from '../../../shared/page-title/page-title';
 import styles from './projects-table.module.css';
-import {validateWBS, WbsNumber} from "utils";
+import { validateWBS, WbsNumber } from 'utils';
 
 export interface DisplayProject {
   wbsNum: string;
@@ -64,7 +64,7 @@ function wbsNumSort(a: string, b: string, order: SortOrder): number {
 /**
  * Interactive table for displaying all projects table data.
  */
-const ProjectsTable: React.FC<DisplayProjectProps> = ({allProjects}: DisplayProjectProps) => {
+const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayProjectProps) => {
   const history = useHistory();
 
   // Configures display options for all data columns
@@ -125,23 +125,23 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({allProjects}: DisplayProj
   };
 
   return (
-      <>
-        <PageTitle title={'All Projects'}/>
-        <BootstrapTable
-            striped
-            hover
-            condensed
-            bootstrap4
-            wrapperClasses={styles.table}
-            keyField="wbsNum"
-            data={allProjects}
-            columns={columns}
-            defaultSorted={defaultSort}
-            rowEvents={rowEvents}
-            noDataIndication="No Projects to Display"
-            rowStyle={{cursor: 'pointer'}}
-        />
-      </>
+    <>
+      <PageTitle title={'All Projects'} />
+      <BootstrapTable
+        striped
+        hover
+        condensed
+        bootstrap4
+        wrapperClasses={styles.table}
+        keyField="wbsNum"
+        data={allProjects}
+        columns={columns}
+        defaultSorted={defaultSort}
+        rowEvents={rowEvents}
+        noDataIndication="No Projects to Display"
+        rowStyle={{ cursor: 'pointer' }}
+      />
+    </>
   );
 };
 
--- a/src/components/projects/projects-table/projects-table/projects-table.tsx
+++ b/src/components/projects/projects-table/projects-table/projects-table.tsx
@@ -25,39 +25,47 @@ interface DisplayProjectProps {
   allProjects: DisplayProject[];
 }
 
-/***
+/**
  * Custom sorting order for wbsNums according to car, then project, then workPackage.
  * @param a 1st wbsNum in string form
  * @param b 2nd wbsNum in string form
  * @param order Imported SortOrder values 'asc' or 'desc'
- * @return number A number -1, 0, or 1 describing the value of a relative to b,
+ * @return number A number describing the value of a relative to b,
  *                according to the specified SortOrder.
  */
-function wbsNumSort(a: string, b: string, order: SortOrder): number {
+export function wbsNumSort(a: string, b: string, order: SortOrder) {
   const wbs_a = validateWBS(a);
   const wbs_b = validateWBS(b);
   if (wbs_a.car !== wbs_b.car) {
     if (order === 'asc') {
-      return wbs_a.car > wbs_b.car ? 1 : -1;
+      return wbs_a.car > wbs_b.car ? wbs_a.car - wbs_b.car : wbs_b.car - wbs_a.car;
     } else {
-      return wbs_a.car > wbs_b.car ? -1 : 1;
+      return wbs_a.car > wbs_b.car ? wbs_b.car - wbs_a.car : wbs_a.car - wbs_b.car;
     }
   }
   if (wbs_a.project !== wbs_b.project) {
     if (order === 'asc') {
-      return wbs_a.project > wbs_b.project ? 1 : -1;
+      return wbs_a.project > wbs_b.project
+        ? wbs_a.project - wbs_b.project
+        : wbs_b.project - wbs_a.project;
     } else {
-      return wbs_a.project > wbs_b.project ? -1 : 1;
+      return wbs_a.project > wbs_b.project
+        ? wbs_b.project - wbs_a.project
+        : wbs_a.project - wbs_b.project;
     }
   }
   if (wbs_a.workPackage !== wbs_b.workPackage) {
     if (order === 'asc') {
-      return wbs_a.workPackage > wbs_b.workPackage ? 1 : -1;
+      return wbs_a.workPackage > wbs_b.workPackage
+        ? wbs_a.workPackage - wbs_b.workPackage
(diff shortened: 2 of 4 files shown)


src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
   // TODO: Sort by wbsNum means 1.1.0 > 1.12.0 > 1.2.0, but desired is 1.1.0 > 1.2.0 > 1.12.0
```
jamescd18, line 34: be sure to remove TODO comments when you've fixed the TODO

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
-    { dataField: 'projectLead', text: 'Project Lead', align: 'left', sort: true },
-    { dataField: 'projectManager', text: 'Project Manager', align: 'left', sort: true },
-    { dataField: 'duration', text: 'Duration', align: 'center', sort: true }
+    {
+      headerAlign: 'center',
+      dataField: 'wbsNum',
+      text: 'WBS #',
+      align: 'center',
+      sort: true,
+      sortFunc: (a, b, order: SortOrder) => {
```
jamescd18, line 43: let's define the sort function inside the component file, but outside the component function, that way we can export it on its own and test it independently in this component's test file.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
-    { dataField: 'wbsNum', text: 'WBS #', align: 'center', sort: true },
-    { dataField: 'name', text: 'Name', align: 'left', sort: true },
-    { dataField: 'projectLead', text: 'Project Lead', align: 'left', sort: true },
-    { dataField: 'projectManager', text: 'Project Manager', align: 'left', sort: true },
-    { dataField: 'duration', text: 'Duration', align: 'center', sort: true }
+    {
+      headerAlign: 'center',
+      dataField: 'wbsNum',
+      text: 'WBS #',
+      align: 'center',
```
jamescd18, line 41: I don't recall the ticket asking for the text within the column to be centered, only the column headers.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
+      sort: true,
+      sortFunc: (a, b, order: SortOrder) => {
+        const a1 = a.split('.');
+        const b1 = b.split('.');
+        const len = Math.min(a1.length, b1.length);
```
jamescd18, line 46: what suggests that the length of a1 would be different from b1?

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
-    { dataField: 'duration', text: 'Duration', align: 'center', sort: true }
+    {
+      headerAlign: 'center',
+      dataField: 'wbsNum',
+      text: 'WBS #',
+      align: 'center',
+      sort: true,
+      sortFunc: (a, b, order: SortOrder) => {
+        const a1 = a.split('.');
+        const b1 = b.split('.');
```
jamescd18, line 45: you appear to be parsing a WBS # string by splitting it on the '.' separator. We [already have a function written](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/85461e78652c3fc564bafd32a935b329f0d2a401/src/utils/src/validate-wbs.ts#L13) that takes in a WBS # string and returns the WBS # in object format adhering to the WBS # interface.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
+        const len = Math.min(a1.length, b1.length);
+        for (let i = 0; i < len; i++) {
+          const a2 = +a1[i] || 0;
```
jamescd18, line 48: avoid using unintelligible variable names. it quickly becomes difficult to trace which variable is what

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
+          const b2 = +b1[i] || 0;
+          if (a2 !== b2) {
+            if (order === 'asc') {
```
jamescd18, line 51: here you're comparing a value which is of the `SortOrder` type against a string. I don't recall specifically what the `SortOrder` type is, but is it not possible to compare against `SortOrder.<SOMETHING>` instead of against a hard-coded string? Hard-coded strings are inherently less flexible

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
+              return a2 > b2 ? 1 : -1;
```
jamescd18, line 52: what is the signature of a sort function? Is it required to return 1 or -1?

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
+            } else {
+              return a2 > b2 ? -1 : 1;
+            }
+          }
+        }
+        return b1.length - a1.length;
```
jamescd18, line 58: how does the length impact the sorting?

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -34,11 +34,52 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
...
+      sort: true
+    },
+    {
+      headerAlign: 'center',
+      dataField: 'projectManager',
+      text: 'Project Manager',
+      align: 'center',
+      sort: true
+    },
+    { headerAlign: 'center', dataField: 'duration', text: 'Duration', align: 'center', sort: true }
```
jamescd18, line 82: any idea why this last one does not format like the rest?

jamescd18 (changes requested):
This is a great first start! You're fulfilling what the ticket asks for, and the sort function probably works correctly. That said, there's definitely room for improvement here. Where applicable, I've provided feedback directly attached to the specific lines it is relevant to. Absolutely use the reply feature to ask clarifying questions or provide additional info. Thank you!

jamescd18:
Also be sure to link the associated issue on the right!

src/components/projects/projects-table/projects-table/projects-table.tsx
shanejpark: I wasn't 100% sure if there could be minor version updates (eg. 1.9 to 1.9.1), so I thought it would be good to account for it.

jamescd18: No worries! WBS #s are not the same as semver. WBS #s are always 3 numbers separated by periods.

jamescd18 (commented):
Looking better! Definitely a few suggestions remaining to fix, and the custom sort function should be defined before the component function definition. **Also, be sure to tag your commits with your ticket number.**

src/components/projects/projects-table/projects-table/projects-table.tsx
Ari-S-123: We've been looking for a SortOrder field for "asc" or "des", but it seems like there aren't any. As it stands, SortOrder can only equal these two strings, so a hard-coded string might be the only option. If you know of any solutions, we would like to know.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -3,14 +3,15 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
-import { useHistory } from 'react-router-dom';
```
jamescd18, line 6: Do you have format on save turned on? This formatting should not be changing.

src/components/projects/projects-table/projects-table/projects-table.test.tsx
```diff
@@ -116,4 +116,27 @@ describe('projects table component', () => {
     const wbsNumsAsc: HTMLElement[] = await screen.findAllByText(wbsRegex);
     expect(wbsNumsAsc.map((ele) => ele.innerHTML)).toEqual(expectedWbsOrder.reverse());
   });
+
+  it('handles sorting and reverse sorting the table by wbsNum', async () => {
```
jamescd18, line 120: How is this different from other tests? For testing, the tests you want to add are ones that are going to just test the sorting function you added. Take a look at this file `src/utils/tests/validate-wbs.test.ts` to see how we're testing a function instead of a component.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -3,14 +3,15 @@
...
+import {useHistory} from 'react-router-dom';
 import BootstrapTable, {
   ColumnDescription,
   RowEventHandlerProps,
   SortOrder
 } from 'react-bootstrap-table-next';
 import PageTitle from '../../../shared/page-title/page-title';
 import styles from './projects-table.module.css';
+import {validateWBS, WbsNumber} from "utils";
```
jamescd18, line 14: This import doesn't look to be formatted quite right.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -24,21 +25,89 @@ interface DisplayProjectProps {
   allProjects: DisplayProject[];
 }
 
+/***
```
jamescd18, line 28: I think this should be two stars, not three

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -24,21 +25,89 @@ interface DisplayProjectProps {
...
+ * Custom sorting order for wbsNums according to car, then project, then workPackage.
+ * @param a 1st wbsNum in string form
+ * @param b 2nd wbsNum in string form
+ * @param order Imported SortOrder values 'asc' or 'desc'
+ * @return number A number -1, 0, or 1 describing the value of a relative to b,
+ *                according to the specified SortOrder.
+ */
+function wbsNumSort(a: string, b: string, order: SortOrder): number {
```
jamescd18, line 36: You might be able to take advantage of type inference and not need to explicitly define the output type.

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -24,21 +25,89 @@ interface DisplayProjectProps {
...
+  const wbs_a = validateWBS(a);
+  const wbs_b = validateWBS(b);
+  if (wbs_a.car !== wbs_b.car) {
+    if (order === 'asc') {
+      return wbs_a.car > wbs_b.car ? 1 : -1;
```
jamescd18, line 41: Does a sort function only return 0, 1, and -1? Or does a sort function return 0, positive number, and negative number?

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -24,21 +25,89 @@ interface DisplayProjectProps {
...
+    }
+  } else {
+    return 0;
+  }
+}
+
 /**
  * Interactive table for displaying all projects table data.
  */
-const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayProjectProps) => {
```
jamescd18, line 30: Formatting should not change

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -56,22 +125,23 @@ const ProjectsTable: React.FC<DisplayProjectProps> = ({ allProjects }: DisplayPr
   };
 
   return (
-    <>
```
jamescd18, line 59: Pretty sure this is also a formatting thing that shouldn't change

jamescd18 (changes requested):
Getting better! Try to pay attention to what lines change, especially if things are changing only because of formatting, make sure the formatting is formatted properly.

src/utils/tests/wbsNumSort.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  it('a.car < b.car, asc', () => {
+    const a = '0.7.12';
+    const b = '1.56.44';
+
+    expect(wbsNumSort(a, b, 'asc')).toEqual(1);
+  });
+
+  it('a.car > b.car, asc', () => {
```
jamescd18, line 44: The test body here matches the previous test but has a different name. Did you intend to change the test body to test something similar but different?

src/utils/tests/wbsNumSort.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  it('a.car < b.car, desc', () => {
+    const a = '0.7.12';
+    const b = '1.56.44';
+
+    expect(wbsNumSort(a, b, 'desc')).toEqual(-1);
+  });
+
+  it('a.car > b.car, desc', () => {
+    const a = '0.7.12';
```
jamescd18, line 59: Same here with test body is same and names are different

src/utils/tests/wbsNumSort.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  it('a.project < b.project, asc', () => {
+    const a = '1.7.12';
+    const b = '1.56.44';
+
+    expect(wbsNumSort(a, b, 'asc')).toEqual(49);
+  });
+
+  it('a.project > b.project, asc', () => {
+    const a = '1.7.12';
```
jamescd18, line 73: Same body vs name

src/components/projects/projects-table/projects-table/projects-table.tsx
```diff
@@ -24,21 +25,97 @@ interface DisplayProjectProps {
...
-    { dataField: 'projectManager', text: 'Project Manager', align: 'left', sort: true },
-    { dataField: 'duration', text: 'Duration', align: 'center', sort: true }
+    {
+      headerAlign: 'center',
+      dataField: 'wbsNum',
+      text: 'WBS #',
+      align: 'center',
+      sort: true,
+      // Custom sort order for wbsNum.
+      sortFunc: (a, b, order: SortOrder) => {
```
jamescd18, line 87: You might be able to just supply the sort function you wrote as the value for sortFunc

jamescd18 (changes requested):
Love the thoroughness with testing, but let's make sure we don't have tons of duplicate tests

src/utils/tests/wbsNumSort.test.ts
Ari-S-123: Yes, some of the tests had an incorrect expected value with the wrong sign. This also revealed a bug within the sorting function which for some reason didn't make itself apparent at all when the app was running.

src/utils/tests/wbsNumSort.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const b = '24.36.12';
+
+    expect(wbsNumSort(a, b, 'desc')).toEqual(0);
+  });
+
+  it('a.car < b.car, asc', () => {
+    const a = '0.7.12';
+    const b = '1.56.44';
+
+    expect(wbsNumSort(a, b, 'asc')).toEqual(-1);
```
jamescd18, line 41: let's inline these strings for simplicity

jamescd18 (changes requested):
Almost there! Just a bit of cleanliness items

src/utils/tests/wbsNumSort.test.ts
```diff
@@ -0,0 +1,72 @@
+/*
```
jamescd18, line 1: I just realized this test file is in the wrong location. Let's keep this test file in the same folder as the component where the function being tested is written.

jamescd18 (changes requested):
Almost there! I love how much cleaner the tests are, but let's be sure the test file is the correct location. Thank you!

jamescd18 (approved):
Awesome! So glad to see it develop and improve along the way!

Issue #332, opened 2021-09-19
Change Requests - Create Change Request Shared Tools


jamescd18:
In order to properly fit the data supplied by the database query to the application, data transformations, and front-end, we need a specific interface for a new and narrower change request summary interface.

The interface should contain:
- change request Id
- submitter
- WBS #
- change request type
- date reviewed
- accepted
- date implemented

Field names should generally match the corresponding field in the database in order to avoid having to rename fields in the backend. 

Dev should branch off of "database-setup" and the PR for this issue should target the "database-setup" branch.

jamescd18:
We are pivoting the direction of this ticket. Instead of creating a separate interface, let's use the same interface and just build variables we can use in the backend across the multiple endpoints.

The shared things are:
- the prisma query include (specifies which fields to fetch from the database)
- the generated type of the output of the prisma query, built using the query input specification (above)
- the transformer function which translates the prisma query output objects into application type objects

This will be more flexible because we won't be limiting which fields we send to the front-end.

PR #341, opened 2021-09-23
#332 - Narrow Change Request Interface

acd124:
We have rewritten the Change Request Interface to be narrower and follow the names of the database schema. We have not implemented any changes to code to follow the new interface.


jamescd18 (changes requested):
I realize now that the ticket was probably not sufficiently clear, but the ticket says "new and narrower interface", which means a new interface should be created. This new interface should be narrower, while the existing change request interface remains unchanged.

If you think a bit about it, you would never change an existing interface that's already in use without extensive changes to test data, interface usages, etc.

src/backend/functions/change-requests.ts
```diff
@@ -4,70 +4,94 @@
...
+): ChangeRequest => {
+  const wbsNum = {
+    car: changeRequest.wbsElement.carNumber,
+    project: changeRequest.wbsElement.projectNumber,
+    workPackage: changeRequest.wbsElement.workPackageNumber
+  };
+  return {
+    id: changeRequest.crId,
+    submitter: changeRequest.submitter,
+    dateSubmitted: changeRequest.dateSubmitted,
```
jamescd18, line 58: These fields are the fields you want to avoid having to write out every single one by simply using the spread operator '...' to dump all fields from the changeRequest object into the object you're constructing.

jamescd18 (changes requested):
Consider using the spread operator to shorten and simplify code. Then also look at how you can simplify it even more by modifying the names of fields in application type interfaces. Feel free to run fields by me before jumping to renaming them because renaming would need to happen in a lot of places to work properly.

jamescd18 (changes requested):
It's looking good! Are you sure you caught all the relevant usages and dependencies for the renamed field? Also noted a few other items

jpwaves:
> It's looking good! Are you sure you caught all the relevant usages and dependencies for the renamed field? Also noted a few other items

I believe we caught all relevant usages of `id` for the `ChangeRequest` type and renamed them to `crId`, but we can go back and double check.

src/components/change-requests/change-requests-table/change-requests-table.tsx
```diff
@@ -23,7 +23,7 @@ const ChangeRequestsTable: React.FC = () => {
   const transformToDisplayChangeRequests = (changeRequests: ChangeRequest[]) => {
     return changeRequests.map((cr: ChangeRequest) => {
       return {
-        id: cr.id,
+        id: cr.crId,
```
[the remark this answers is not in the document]
jpwaves: We have not updated the `DisplayChangeRequest` interface; so far, we've only modified the `ChangeRequest` interface.Would you like us to rename the `id` property in the `DisplayChangeRequest` type to `crId` and simplify the return object with the spread operator like we did with the `ChangeRequest` type?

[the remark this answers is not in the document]
jamescd18: Oh oopsies I think it looks good as-is

src/components/projects/wbs-details/work-package-container/changes-list/changes-list.tsx
```diff
@@ -20,7 +20,8 @@ const ChangesList: React.FC<ChangesListProps> = ({ changes }) => {
       headerRight={<></>}
       list={changes.map((ic) => (
         <>
-          [<Link to={`${routes.CHANGE_REQUESTS}/${ic.crId}`}>#{ic.crId}</Link>] {ic.detail}
+          [<Link to={`${routes.CHANGE_REQUESTS}/${ic.changeRequestId}`}>#{ic.changeRequestId}</Link>
+          ] {ic.detail}
```
jamescd18, line 24: The formatting of the code here looks a little funky. Is it what prettier did? Or can it be adjusted with some manual formatting, like split the two tags for the link element onto their own lines. And if we do manual formatting does it accidentally change the display on the front-end or no?

jamescd18 (commented):
This is looking really good! There's a couple things I'd love to verify and double check, but otherwise it should be ready to merge before Wednesday. Are there any outstanding tasks for this ticket?

src/components/projects/wbs-details/work-package-container/changes-list/changes-list.tsx
acd124: It is what prettier did, and does not change any of the display on the front end that I could tell. I tried a few manual changes with more new lines and it still did not change the display unless I added non-whitespace characters. Did you want something like this with many shorter lines? I'm not sure if the indentation is correct since all prettier would do given the current settings is revert it to what it currently is.
Screen Shot 2021-10-04 at 9 10 36 PM

Issue #317, opened 2021-08-24
General - Create Glossary


jamescd18:
Given how most users of the application are unfamiliar with Project Management or the concepts and processes facilitated by the application, it would be quite useful to include a glossary of terms in the application. PM-Dashboard-v1 had a small glossary.

PR #348, opened 2021-09-29
#317 create glossary front end

baydeni:
image


jamescd18 (changes requested):
This is a great start! The page block is used properly and tested well. You'll need to remove the .env file, and also try not to include lots of extraneous new lines.

baydeni:
image

Issue #347, opened 2021-09-27
General - Explicitly Type Caught Errors


acd124:
Make the any type for errors caught in try catch blocks explicit to resolve the Typescript noImplicitAny error.

PR #350, opened 2021-09-29
#347 Specified type to any for errors in catch blocks under the funct…

Ari-S-123:
We specified the error type to any in all four try-catch blocks under the functions package in the backend.

### DIFF
--- a/src/backend/functions/change-requests.ts
+++ b/src/backend/functions/change-requests.ts
@@ -49,7 +49,7 @@ const routes: ApiRoute[] = [
 const handler: Handler = async (event, context) => {
   try {
     return routeMatcher(routes, event, context);
-  } catch (error) {
+  } catch (error: any) {
     console.error(error);
     return buildServerFailureResponse(error.message);
   }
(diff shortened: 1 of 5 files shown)


src/backend/functions/change-requests.ts
```diff
@@ -49,7 +49,7 @@ const routes: ApiRoute[] = [
 const handler: Handler = async (event, context) => {
   try {
     return routeMatcher(routes, event, context);
-  } catch (error) {
+  } catch (error: any) {
```
jamescd18, line 52: I believe there's an `Error` type

jamescd18 (changes requested):
Try not 'any'

src/backend/functions/change-requests.ts
Ari-S-123: Screenshot (979)
If I try that then this happens. It seems like it has to be either 'any' or 'unknown'.

KincentLan: @jamescd18 https://stackoverflow.com/questions/42618089/how-do-you-use-typed-errors-in-async-catch
TLDR; catch clause variables may not have a type annotation

jamescd18: Thank you for clarifying! Let's go with just `any` then?

Issue #119, opened 2021-03-26
General - Add Tests to Shared Folder


jamescd18:
Currently, there is no test coverage for any of the files in the `src/shared` folder. This is potentially problematic because a small basic level of tests helps us catch bugs faster and prevent accidentally breaking the code. Let's add a tests folder and two test files.

- [x] Create a new folder `/src/shared/tests`
- [x] Add a test file to correspond to the `pipes.tsx` file
  - [x] Add a `describe()` test suite to correspond to each of the functions in the `pipes.tsx` file
  - [x] Add one or more `it()` test methods to each test suite to have a minimal baseline of tests
- [ ] Add a test file to correspond to the `test-utils.tsx` file
  - [ ] Add a `describe()` test suite to correspond to each of the functions in the `test-utils.tsx` file
  - [ ] Add one or more `it()` test methods to each test suite to have a minimal baseline of tests

PR #353, opened 2021-10-05
#119 - Shared Folder Tests

Ari-S-123:
We wrote tests for all functions in pipes.test.tsx. We did not write any tests for the functions in test-utils.tsx as it was moved out of the shared folder.


src/shared/tests/pipes.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+describe('Formatting Dates', () => {
+  test('invalid test', () => {
+    expect(datePipe(new Date(99999999999999999999))).toBe('Invalid Date');
+  });
+  test('empty constructor/current time', () => {
+    let current_time = new Date();
+    let answer = current_time.toLocaleString('en-US');
+    expect(datePipe(new Date())).toBe(answer);
+  });
+  // This test fails on github but passes locally.
```
jamescd18, line 212: If this is still true, let's just remove it, otherwise, remove the comment

src/shared/tests/pipes.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+  });
+  test('non-empty string with numbers', () => {
+    expect(emptyStringPipe('123')).toBe('123');
+  });
+  test('non-empty string with special characters', () => {
+    expect(emptyStringPipe('!@#$%^&*()}{:">?')).toBe('!@#$%^&*()}{:">?');
+  });
+});
+
+describe('Formatting Empty Strings With Dash', () => {
```
jamescd18, line 185: Do you have any cases here that expect an em dash output?

src/shared/tests/pipes.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+  test('empty string', () => {
+    expect(emDashPipe('')).toBe('');
+  });
+  test('empty string with space', () => {
+    expect(emDashPipe(' ')).toBe(' ');
```
jamescd18, line 190: I think it'd be good to have this behave by outputting an em dash. Check out the trim method if you're not already familiar with it

src/shared/tests/pipes.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+    expect(datePipe(new Date('2021-10-5'))).toBe('10/5/2021, 12:00:00 AM');
+  });
+  test('custom detailed', () => {
+    expect(datePipe(new Date(2021, 0o11, 0o5, 14, 3, 6, 9))).toBe('10/5/2021, 2:03:06 PM');
+  });
+});
+
+describe('Formatting Wbs Numbers', () => {
+  test('with dummy data', () => {
+    expect(wbsPipe(exampleWorkPackage1.wbsNum)).toBe('1.1.1');
```
jamescd18, line 226: I think you can find example WBS#s directly in the stub data rather than pulling from projects/work packages if you want to

src/shared/tests/pipes.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+  });
+  test('datestring', () => {
+    expect(datePipe(new Date('2021-10-5'))).toBe('10/5/2021, 12:00:00 AM');
+  });
+  test('custom detailed', () => {
+    expect(datePipe(new Date(2021, 0o11, 0o5, 14, 3, 6, 9))).toBe('10/5/2021, 2:03:06 PM');
+  });
+});
+
+describe('Formatting Wbs Numbers', () => {
```
jamescd18, line 224: Let's add a test case that doesn't use the dummy data so we can have 2 and 3 digit numbers in the WBS#

src/shared/tests/pipes.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+    expect(weeksPipe(1.5)).toBe('1.5 weeks');
+  });
+  test('with hexadecimal', () => {
+    expect(weeksPipe(0x37cf)).toBe('14287 weeks');
+  });
+  test('with octal', () => {
+    expect(weeksPipe(0o377)).toBe('255 weeks');
+  });
+  test('with binary', () => {
+    expect(weeksPipe(0b111001)).toBe('57 weeks');
```
jamescd18, line 131: I love this, and it's kinda funny cuz it's valid, but this is the kinda test case that would probably be a little much on co-op. 

You can leave it in here, but typically you want to bias towards the minimal number of tests which cover a couple of each base cases and one for each major edge case and no more. This is because more tests = more work to maintain them / edit them when the application changes.

Issue #302, opened 2021-08-16
General - Create Separated Type Transformer Functions


jamescd18:
Create a distinct folder structure for defining type transformer functions. Type transformer functions serve the purpose of ensuring data pulled in by API requests has been properly transformed from JSON string data into proper data structures. Particularly for date strings converted to date objects.

PR #360, opened 2021-10-16
#302 moved change request transformer to separate transformer file

baydeni:
We created a transformers folder and moved the change request transformer into it. We're not sure if we needed to do any tests for it because there weren't any previous tests for it. We are thinking about putting all transformer functions into one file.

### DIFF
--- a/src/services/change-requests.api.ts
+++ b/src/services/change-requests.api.ts
@@ -4,32 +4,9 @@
  */
 
 import axios from 'axios';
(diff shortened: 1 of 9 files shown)


baydeni:
We moved all transformer functions within services into their own folder with a better naming scheme. We aren't sure if there is anything more to do/abstract, and we also aren't sure if there is anything to do with the hooks.

jamescd18 (approved):
Looks great! Simple and consistent structures that do the job which is exactly what we need. Thank you!

Issue #154, opened 2021-04-19
Configs - Update ESLint Configuration


jamescd18:
I happened to discover a couple nice ESLint configs that I'd like us to add to the package.json. If you happen to come across additional useful ESLint configs, feel free to drop them in a comment reply with a quick explanation of why, and we'll consider adding it too.

Confirmed to add:
```json
"rules": {
      "guard-for-in": "error",
      "prefer-arrow-callback": "error"
    },
```

acd124:
possible rules to look at:

- [ ] [no-unsafe-negation](https://eslint.org/docs/rules/no-unsafe-negation): ["error", { "enforceForOrderingRelations": true }]
- [ ] [no-unsafe-optional-chaining](https://eslint.org/docs/rules/no-unsafe-optional-chaining): ["error", { "disallowArithmeticOperators": true }]
- [ ] [arrow-body-style](https://eslint.org/docs/rules/arrow-body-style): ["error", "as-needed"]
- [ ] [block-scoped-var](https://eslint.org/docs/rules/block-scoped-var)
- [ ] [eqeqeq](https://eslint.org/docs/rules/eqeqeq)
- [ ] [max-depth](https://eslint.org/docs/rules/max-depth)
- [ ] [max-nested-callbacks](https://eslint.org/docs/rules/max-nested-callbacks)
- [ ] [no-else-return](https://eslint.org/docs/rules/no-else-return)
- [ ] [no-lonely-if](https://eslint.org/docs/rules/no-lonely-if)
- [ ] [no-throw-literal](https://eslint.org/docs/rules/no-throw-literal)
- [ ] [prefer-promise-reject-errors](https://eslint.org/docs/rules/prefer-promise-reject-errors)
- [ ] [no-unneeded-ternary](https://eslint.org/docs/rules/no-unneeded-ternary)
- [ ] [no-var](https://eslint.org/docs/rules/no-var)
- [ ] [object-shorthand](https://eslint.org/docs/rules/object-shorthand)
- [ ] [operator-assignment](https://eslint.org/docs/rules/operator-assignment)
- [ ] [prefer-const](https://eslint.org/docs/rules/prefer-const)
- [ ] [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring)
- [ ] [prefer-exponentiation-operator](https://eslint.org/docs/rules/prefer-exponentiation-operator)
[...]

PR #386, opened 2021-11-14
#154 - Implement eslint configs

RChandler234:
We implemented most of the eslint configs. We tried to add the prefer-nullish-coallescing setting, but we couldn't get it to work.


jamescd18 (approved):
This looks great! I love keeping code clean and tidy! No worries about the one you weren't able to implement. Thank you!

Issue #358, opened 2021-10-13
Users - Create Users Shared Tools


jamescd18:
In order to properly fit the data supplied by the database query to the application, data transformations, and front-end, we need a shared query and response transformer set of tools in the backend for use in all of the backend users endpoint routes.

See #332, #333, #357, and their associated pull requests for details.

Dev should branch off of "database-setup" and the PR for this issue should target the "database-setup" branch.

PR #387, opened 2021-11-16
#358 create users shared tools

anthonybernardi:
added the user transformer in users.ts


jamescd18 (approved):
Looks great! Love to get this completed

Issue #383, opened 2021-11-10
General - Test `database-setup` Branch


jamescd18:
## Background:
The `database-setup` branch was created in order to implement the messy and sweeping changes it took to properly configure and set up the back-end to query data from the database.

## Description:
Prior to merging the `database-setup` branch into `main`, we need to be confident that the application, as run from the `database-setup` branch, works properly in its entirety. Run the application locally and thoroughly run various manual tests of your choosing. Document the manual tests you performed in a bullet list in a comment reply on this ticket. Be sure to specify any odd behaviors or broken features you encounter to review with the team.

acd124:
Viewing specific Change Requests usually does not work, often erroring with `cannot read property map of undefined` from `src/components/change-requests/change-request-details/change-request-details/change-request-details.tsx:46`. This is likely because of some missing implementation for specific change request types. The only change request that looks to work has crId 5.

RChandler234:
Here are the notes I had after testing the ‘database-setup’ branch; I didn’t find any problems that I felt immediately needed to be addressed, mostly just weird formatting things, let me know if there’s anything else I should test/ if I should address any of the minor issues I mentioned:

-Loading into home page as admin works fine
-All the links on project details link to rick astely (assuming that’s a good thing)
-Not sure what the active button with little arrows next to it is for, just moves the text slightly (same with the rest of the little arrow things)
this thing:
<img width="14" alt="button" src="https://user-images.githubusercontent.com/29521172/147511901-1ed6786c-f5df-4171-a42c-97fbee11ef0e.png">
-Trying to open the change request link in a project doesn’t work if I try to open it in a new tab, but otherwise it works fine (same for work package links)
-Workpackage info dropdown within project pages works well
-Workpackage links work
-reloading the page always makes you sign back in as admin
-The implemented changes tab currently just has the placeholder (“list of changes”) might be nice to see what the format for this would actually look like

-All the projects and subsequent links work in the projects tab,I went through testing that each link within each project worked as expected, checked generally to make sure the content looked correct, and trying some weird things like opening in a new tab, etc.




-Weird format inconsistencies in some change requests: (seems to specifically be in the “new function” type)
Most Change Requests:

<img width="908" alt="normal" src="https://user-images.githubusercontent.com/29521172/147511753-300abb21-d2dd-431b-bcf6-c9035b678bd8.png">

“New Function” change request :
[...]

PR #393, opened 2021-12-11
#332 Add Backend Change Request Types Support

acd124:
Allow change request endpoints to properly supply information to follow types for activation, stage gate, and standard scope change requests. Should take care of the bug from #383.


jamescd18 (approved):
Looks good to me!

Issue #388, opened 2021-11-18
General - Implement custom bootstrap elements using Sass


shanejpark:
## Description

Change usage for colors for the dashboard to implement NER's official colors into bootstrap.

## Background

Currently, in the project, there are multiple areas where there are manual deviations relating to color. The first can either be through complete deviation from our standard colors, or from overriding colors and custom colors.

## Proposed Solution

- [ ] Create the file  **src/custom.scss** that implements NER's official colors into bootstrap
- [ ] Find every bootstrap element that either a) does not follow the official color guidelines, or b) uses !important or some other override to force a color, and properly utilize the custom colors in **src/custom.scss**
- [ ] Change bootstrap imports to only import elements of bootstrap that they utilize

PR #394, opened 2021-12-16
dark mode first attempt, mostly just the context, also settings page

jamescd18:
Very half-assed initial stab at getting dark mode working. I think we should not pursue this until after #388 gets closed and we discuss the approach a bit. We'll definitely need design support on this so we don't end up making a shitty ugly dark mode. I suspect the context I've implemented here will stay, since it offers flexibility to implement a lot of other global client-side settings, which could later be implemented as saved settings by linking them to the back-end. Also I made adjustments to the settings page.

Here's the new settings page
Screen Shot 2021-12-16 at 3 04 12 AM
 

Here's the dark mode attempt:
Screen Shot 2021-12-16 at 3 04 56 AM
Screen Shot 2021-12-16 at 3 05 37 AM

### DIFF
--- a/src/components/app/app-context-settings/app-context-settings.module.css
+++ b/src/components/app/app-context-settings/app-context-settings.module.css
@@ -1,0 +1,4 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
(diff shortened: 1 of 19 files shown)


jamescd18:
A lot more design work is needed, and probably something like CSS variables (#388) or maybe even switching away from bootstrap to tailwind CSS?

Issue #343, opened 2021-09-27
General - Perform production database migration & setup


KincentLan:
## Description

Perform various tasks to prepare the production database for testing.

## Prerequisites

The create and implement change request epics must be completed first
- [x] #374
- [x] #397  

## Proposed solution

- [ ] Export all data from the production database
- [ ] Run the Prisma database migrations on the production database
- [ ] Modify exported data into import-ready script by removing table modification commands and re-ordering insert commands to ensure proper order of dependent tables
- [ ] Run data import script to re-load the database
- [ ] Configure the database user
- [ ] Document the process

## Other Notes

Following the principle of least privilege, we want to make sure that normal application users cannot modify tables that they are not normally supposed to. Normal application users should only be able to read and create, not delete or modify.

we want to load some data into the production database so we can test the production environment to make sure everything is running smoothly.

Reach out to @jamescd18 for pair-programming on this issue (as you need proper privileges to modify the production database).

jinhyunh:
https://docs.google.com/document/d/1y1hMYbGLkjrgl3DwfgE6dtVjzpqiDjN-zltIpPz2VEU/edit?usp=sharing

PR #396, opened 2021-12-21
#343-performed-database-migration

rishavsarma5:
Screenshot (340)

https://docs.google.com/document/d/1y1hMYbGLkjrgl3DwfgE6dtVjzpqiDjN-zltIpPz2VEU/edit?usp=sharing


.gitignore
```diff
@@ -2,8 +2,6 @@
 
 .eslintcache
 
-.env
```
jamescd18, line 5: No. What does the .gitignore do?

.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL="postgresql://postgres:Xboxnam101@localhost:5432/nerpm?schema=public"
```
jamescd18, line 1: This should not be in the codebase. You have exposed your local database password to the public.

package.json
```diff
@@ -189,7 +189,7 @@
     "netlify-lambda": "^2.0.3",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.2.1",
-    "prisma": "~2.19.0",
+    "prisma": "^2.19.0",
```
jamescd18, line 192: You should not be changing versions without a very strong reason

src/backend/prisma/migrations/20211221014206_test/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE "WBS_Element" DROP COLUMN "projectLeadId",
+ADD COLUMN     "projectLeadId" INTEGER[];
+
+-- CreateTable
+CREATE TABLE "_projectLead" (
+    "A" INTEGER NOT NULL,
+    "B" INTEGER NOT NULL
+);
+
+-- CreateIndex
```
jamescd18, line 17: We don't want any changes to the migrations to get added to the codebase.

jamescd18 (commented):
This needs some more explanation please. I see you've made changes to the codebase and linked a google doc. 

If you want to use a Google doc to iterate on the instructions before putting them into the proper codebase docs files, I'm okay with that. That said, at a bare minimum you're going to need to give me edit access so I can leave feedback, and you also should place the file in the correct NER Google Drive folder. That said, I don't recommend using a Google doc since the docs properly live in the codebase, and GitHub provides a great line-by-line feedback tool via pull requests. 

As for all of the changes that I see here in the codebase, they are all incorrect. I've specified why in inline comments, but I would strongly recommend doing a little research into what the gitignore does and why we don't put the .env in the codebase. If you aren't able to find answers, let me know and I can explain.

.gitignore
jinhyunh: It chooses which files to ignore committing the project onto the repository

PR #414, opened 2022-01-02
Create CR endpoint with Middy.js

jamescd18:
Okay so this one is kinda interesting. I'm learning just how little I know / knew about AWS Lambda. I'm also realizing how long a single create endpoint can get compared to the dinky simple get endpoints we started with. It really shows how naive I was in thinking that endpoints would be small enough to bundle together with route matching. 

This honestly probably is the better way to architect our backend. It just feels so much cleaner. And probably more testable honestly, though I haven't given that a shot yet. 

This likely also puts us in a much better spot for a future in which we might consider migrating to full-fledged AWS instead of riding off of Netlify, but that future scares me given how little DevSecOps knowledge our team has at the moment.


src/backend/functions/change-request-new.ts
```diff
@@ -0,0 +1,102 @@
...
+      break;
+  }
+  // TODO: change this return statement
+  return buildSuccessResponse(created || { result: 'success', message: 'cr created correctly' });
+};
+
+// validates the event parameter, so body must be explicitly stated
+// TODO: actually complete the schema
+// TODO: consider moving schemas (or parts of schemas) to utils package? consider using schema-to-ts?
+const inputSchema = {
```
jamescd18, line 82: Note: the validation schema I have here is definitely a work-in-progress. I was examining how to build the complex typing of a change request into JSON schema and simply didn't have enough time to figure it out since I wanted to move forward with this PR. Here's some of what I was looking at:

https://stackoverflow.com/questions/9029524/json-schema-specify-field-is-required-based-on-value-of-another-field

src/backend/functions/change-request-new.ts
```diff
@@ -0,0 +1,102 @@
...
+      },
+      required: ['submitterId', 'wbsElementId', 'type']
+    }
+  }
+};
+
+const handler = middy(baseHandler)
+  .use(jsonBodyParser())
+  .use(validator({ inputSchema }))
+  .use(httpErrorHandler());
```
jamescd18, line 100: Note: this is just the example Middy.js provides. There may very well be other middleware that is useful for us, and Middy.js does allow writing your own middleware too, I just haven't explored any of that or any further configuration of these middleware.

src/backend/functions/change-request-new.ts
jamescd18: Currently, bad input values seems to displease Prisma in ways I don't fully understand, so I think it's likely wisest for us to be validating all the inputs. I also think this syntax is kinda nice for defining input constraints.

jamescd18:
Okay so I think this PR is actually basically in a ready state to merge if we wanted to, but here's the loose ends:
- [ ] Testing for this new design pattern / architecture
- [ ] Updating back-end route strings (aka re-linking the front-end to this new back-end design)

As soon as we moved away from hard-coded data in the back-end to linking to the database, we basically stopped testing the back-end because I couldn't get proper test mocking to work to remove the Prisma client. I haven't thought about it yet, but I'm kinda guessing that this new pattern may be easier to test.

acd124 (commented):
I like the way this is looking, there's certainly a lot more to look into on this front, but validation looks a lot simpler now.

src/backend/functions/change-request-new.ts
jamescd18: Your example of `GET /change-requests/word` refers to a different endpoint though, no?

src/backend/functions/change-request-new.ts
jamescd18: 100% agree. Authorization would be the next middleware to tackle.

src/backend/functions/change-request-new.ts
acd124: Yes, it's a different endpoint, but reinforcing that all user sent requests need to be validated

jamescd18:
Update: I did a test deploy and the Netlify Deploy Preview build failed so I put in a request to ITS to request those logs so we can determine what's wrong. It's possible it's just a lint error, or it could be something worse.

jamescd18:
As recommended by @acd124, it looks like changing to the new `esbuild` node bundler in the netlify.toml fixed the failing deploy preview.

https://answers.netlify.com/t/error-could-not-find-encoding-module-in-file-netlify-function/2259/19

jamescd18:
yet now I get another error that says utils js file exports is not found. I suspect it might have something to do with the new node bundler since that enables esmodule syntax natively supported by the functions when originally we set up the utils package (for usage with the functions) to use commonJS module syntax. alas, this switch did not fix it so I am awaiting ITS' response for logs again

Screen Shot 2022-01-10 at 5 26 42 PM
Screen Shot 2022-01-10 at 5 26 56 PM

acd124:
https://answers.netlify.com/t/lambda-function-es-module-support/30673/16 The answer to the problem is likely somewhere in this thread, though we'll have to wait for the logs to confirm what the right solution is. I believe the final messages in the thread say that using esbuild bundler does work, but we may want to specify using node 14.

willfarrell:
Hi, hope you're enjoying using Middy. Saw you linked to a middy issues, so thought I'd poke around. Wanted to let you know that v3 will have a `http-router` if that helps (https://github.com/middyjs/middy/pull/782). Also it might be worth checking out https://github.com/willfarrell/middy-ajv if you plan to have a lot of cold starts.

acd124:
Following research into https://answers.netlify.com/t/local-npm-dependency-in-package-json-not-installing/22291/14 the current best solution for making sure utils is properly installed is to include the preinstall script in the build script. Looking into the underlying cause shows that it is an issue with netlify's caching and there is an issue open https://github.com/netlify/build-image/issues/523 about it. There is also a solution pending that could fix it as shown in https://github.com/netlify/build-image/pull/513 more research might be done to see if there are useful tools or changes to be learned here about our deployment process, but for now the main goal is to get the app running.

jamescd18:
Thank you @willfarrell! Really excited to see v3 come out!

jamescd18:
Okay! I looks like adding `npm ci` to run before the build script fixed it. Now we just need to get the login page Google login button to work properly too, so I'm waiting for ITS to rename the env variable.

jamescd18:
The front-end doesn't work because of the env variable naming, but I can still send requests to the back-end via Thunder Client (VSCode integrated Postman). Sending requests to this new endpoint properly trigger "event object failed validation" so I'd like to believe that it's actually fully working. Obviously we still need to do stuff for the production database and stuff, but this is a huge step forward.

jamescd18:
SUCCESSFUL EXECUTION OF THE CREATE ENDPOINT!!!!!!!!!

Screen Shot 2022-01-13 at 3 15 20 AM

Screen Shot 2022-01-13 at 3 16 02 AM

Issue #418, opened 2022-01-04
General - Page Title Scroll on Overflow for Action Button


jpwaves:
### Browser(s)

Firefox

### Platform(s)

Desktop

### Version(s)

main

### Observed Behavior

When a component is added to a `PageTitle` in the `actionButton` prop, the resulting component sometimes renders with and horizontal and/or vertical scrollbar for the component due to overflow despite the button being visually fine. This happens with single buttons, a row of buttons, and a dropdown button. 

### Expected Behavior

For single buttons and a row of buttons, there shouldn't need to be any overflow either horizontally or vertically. For a dropdown button, the dropdown menu should either overlap or push down on any components below it on the screen.

### Steps to Reproduce

1. Create a button or action button with JSX within the component using a `PageTitle`
2. Add that button/action button to `PageTitle` via the `actionButton` prop
3. Launch the application and navigate to the page that you modified

### Screenshots (as needed)

Dropdown button without menu extended
![default](https://user-images.githubusercontent.com/48561685/148020234-faef194c-60b8-43b4-8936-7f84cd79ef3f.png)
Dropdown button with menu extended (creates a vertical overflow that causes the vertical scrollbar to be added)
![scrolled](https://user-images.githubusercontent.com/48561685/148020313-47a08f3b-07b1-4ab8-962d-b3c3fe4124a1.png)
![unscrolled](https://user-images.githubusercontent.com/48561685/148020321-72388b66-34ca-41df-abc3-c26abdefca82.png)
Row of buttons (horizontal overflow)
![horizontal-btns-overflow](https://user-images.githubusercontent.com/48561685/148020410-c7b47333-09be-4801-a245-5f06d43da85f.png)
Single button (horizontal overflow)
![horizontal-single-btn](https://user-images.githubusercontent.com/48561685/148020559-6cb50cdf-cc6c-4b3e-895e-18aec44e5926.png)

Issue #354, opened 2021-10-06
Change Requests - Create Review Visuals


rswernofsky:
## Description
Two new components need to be created so that project managers can review change requests. One component is the action dropdown where project managers can choose to either accept or deny a change request. The other components are the accept/deny modals that appear when reviewing a change request that allow project managers to add review notes. 

## Background
The idea is that project managers will click on the action dropdown on individual change requests and choose either the accept or deny option. Depending on which option is chosen, a modal will be appear on the screen prompting the project managers to add review notes and confirm their decision. This ticket is to create the action dropdown component and the accept and deny modals.

### Action dropdown 
- [ ] Create dropdown that has placeholder text "Action" with two options: "Accept" and "Deny"

### Review Modal
The two modals are very similar — they only differ in the headings. Therefore, it is possible to create one component and use conditionals to display the different headings. 
- [ ] Create modal that conditionally displays "Accept Change Request" or "Deny Change Request" as heading, depending on option that user chose in dropdown
- [ ] There should be a free text section with label "Review Notes"
- [ ] Cancel and confirm button should be present on bottom right corner of modal

## Mocks
https://www.figma.com/file/qmXuySSwfiD942HwmSSEMe/NERPM-Desktop?node-id=201%3A549

PR #420, opened 2022-01-08
#354 - Create cr review component and buttons to navigate to it

jpwaves:
I created the component for reviewing change requests that conditionally shows either "Accept Change Request" or "Deny Change Request" depending on which action button was used to navigate to the page. Currently, the "Cancel" button redirects the user to the details page for the particular change request and the "Confirm" button redirects to the page with all the change requests (I assumed that's the expected behavior after successfully submitting a change request). 

I added two action buttons instead of a dropdown menu due to a problem with the `PageTitle` component that is explained in #418. 

Side note, for some of the tests, it seems that adding the two action buttons to the change requests details page causes a number of tests for that component to fail (and it already seems to have some tests failing even before my changes). The error message states: `Invariant failed: You should not use <Link> outside a <Router>`. I'm not sure if this is something I should address in the changes I made to the details page, but I'd like to any insight on that issue (or if it's something I should just ignore).

Screenshots:
Action buttons for navigating to the review change request page
action-btns

Review change request page
accept-cr

deny-cr


src/components/change-requests/change-request-details/change-request-details/change-request-details.module.css
```diff
@@ -6,3 +6,8 @@
 b {
   font-weight: bold;
 }
+
+.btnsContainer {
+  display: flex;
+  gap: 1.5rem;
+}
```
jamescd18, line 13:
```suggestion
}

```

src/components/change-requests/change-requests.tsx
```diff
@@ -8,11 +8,18 @@ import { routes } from '../../shared/routes';
...
 import ChangeRequestDetails from './change-request-details/change-request-details';
 import NewChangeRequest from './new-change-request/new-change-request';
+import ReviewChangeRequest from './review-change-request/review-change-request';
 import './change-requests.module.css';
 
 const ChangeRequests: React.FC = () => {
   return (
     <Switch>
+      <Route path={routes.CHANGE_REQUESTS_ACCEPT}>
+        <ReviewChangeRequest accepted={true} />
```
jamescd18, line 18:
```suggestion
        <ReviewChangeRequest accepted />
```
There's actually a shorthand where this ^ is equivalent to the code you wrote https://stackoverflow.com/questions/60816731/how-are-boolean-props-used-in-react

src/components/change-requests/change-requests.tsx
```diff
@@ -8,11 +8,18 @@ import { routes } from '../../shared/routes';
...
+      </Route>
+      <Route path={routes.CHANGE_REQUESTS_DENY}>
+        <ReviewChangeRequest accepted={false} />
```
jamescd18, line 21:
```suggestion
        <ReviewChangeRequest accept />
      </Route>
      <Route path={routes.CHANGE_REQUESTS_DENY}>
        <ReviewChangeRequest accept={false} />
```

I would also suggest renaming the prop so it's a little bit clearer since the component provides the user with the _option_ to accept the CR

src/components/change-requests/review-change-request/review-change-request.module.css
```diff
@@ -0,0 +1,4 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
```
jamescd18, line 4:
```suggestion
 */
 
```

src/components/change-requests/review-change-request/review-change-request.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+        expect(screen.getByText('Accept Change Request')).toBeInTheDocument();
+    });
+
+    it('renders change request review for denying', () => {
+        renderComponent(false);
+
+        expect(screen.getByText('Deny Change Request')).toBeInTheDocument();
+    });
+});
```
jamescd18, line 32:
```suggestion
});

```

src/components/change-requests/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,24 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import { useParams } from "react-router-dom";
+import ReviewChangeRequestsView from "./review-change-request/review-change-request";
+
+interface ReviewChangeRequestProps {
+    accepted: boolean
```
jamescd18, line 10:
```suggestion
    accept: boolean
```

src/components/change-requests/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,24 @@
...
+}: ReviewChangeRequestProps) => {
+    interface ParamTypes {
+        id: string;
+    }
+    const { id } = useParams<ParamTypes>();
+
+    return <ReviewChangeRequestsView crId={parseInt(id)} accepted={accepted} />;
+};
+
+export default ReviewChangeRequest;
```
jamescd18, line 24:
```suggestion
export default ReviewChangeRequest;

```

src/components/change-requests/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const ReviewChangeRequest: React.FC<ReviewChangeRequestProps> = ({
+    accepted
+}: ReviewChangeRequestProps) => {
+    interface ParamTypes {
+        id: string;
+    }
+    const { id } = useParams<ParamTypes>();
+
+    return <ReviewChangeRequestsView crId={parseInt(id)} accepted={accepted} />;
```
jamescd18, line 21:
```suggestion
    return <ReviewChangeRequestsView crId={parseInt(id)} accept={accept} />;
```

src/components/change-requests/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { useParams } from "react-router-dom";
+import ReviewChangeRequestsView from "./review-change-request/review-change-request";
+
+interface ReviewChangeRequestProps {
+    accepted: boolean
+}
+
+const ReviewChangeRequest: React.FC<ReviewChangeRequestProps> = ({
+    accepted
```
jamescd18, line 14:
```suggestion
    accept
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.module.css
```diff
@@ -0,0 +1,6 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+
```
jamescd18, line 6:
```suggestion
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { exampleStandardChangeRequest } from "../../../../test-support/test-data/change-requests.stub";
+import { render, screen } from "@testing-library/react";
+import ReviewChangeRequestsView from "./review-change-request";
+
+/**
+ * Sets up the component under test with the desired values and renders it.
+ */
+const renderComponent = (accepted: boolean) => {
+    return render(<ReviewChangeRequestsView crId={exampleStandardChangeRequest.crId} accepted={accepted} />);
```
jamescd18, line 14:
```suggestion
    return render(<ReviewChangeRequestsView crId={exampleStandardChangeRequest.crId} accept={accept} />);
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import { exampleStandardChangeRequest } from "../../../../test-support/test-data/change-requests.stub";
+import { render, screen } from "@testing-library/react";
+import ReviewChangeRequestsView from "./review-change-request";
+
+/**
+ * Sets up the component under test with the desired values and renders it.
+ */
+const renderComponent = (accepted: boolean) => {
```
jamescd18, line 13:
```suggestion
const renderComponent = (accept: boolean) => {
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.test.tsx
```diff
@@ -0,0 +1,48 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import { exampleStandardChangeRequest } from "../../../../test-support/test-data/change-requests.stub";
+import { render, screen } from "@testing-library/react";
```
jamescd18, line 7: you'll actually need to import from one of our utils files because we did some configuring to modify the render function

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1121297ac2263c7c55085c8faef3fd2a759f0ed7/src/components/change-requests/change-request-details/change-request-details.test.tsx#L9

src/components/change-requests/review-change-request/review-change-request/review-change-request.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+        expect(screen.getByPlaceholderText('Notes...')).toBeInTheDocument();
+    });
+
+    it('renders buttons', () => {
+        renderComponent(true);
+
+        expect(screen.getByText('Cancel')).toBeInTheDocument();
+        expect(screen.getByText('Confirm')).toBeInTheDocument();
+    });
+});
```
jamescd18, line 48:
```suggestion
});

```

src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,73 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+import { Button, Form } from 'react-bootstrap';
+import { useHistory } from 'react-router-dom';
+import {
+  ChangeRequest,
```
jamescd18, line 8: make sure to remove any unused imports!

src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,73 @@
...
+  StageGateChangeRequest,
+  ChangeRequestType,
+  ChangeRequestExplanation
+} from 'utils';
+import PageBlock from "../../../shared/page-block/page-block";
+import PageTitle from "../../../shared/page-title/page-title";
+
+interface ReviewChangeRequestViewProps {
+  crId: number,
+  accepted: boolean
```
jamescd18, line 20:
```suggestion
  accept: boolean
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+const ReviewChangeRequestsView: React.FC<ReviewChangeRequestViewProps> = ({
+  crId,
+  accepted
```
jamescd18, line 25:
```suggestion
  accept
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,73 @@
...
+}: ReviewChangeRequestViewProps) => {
+  const history = useHistory();
+
+  const handleConfirm = () => {
+    // temporary redirect to change requests table
```
jamescd18, line 30: I actually like this because the "better" way would be a tad more complicated than I think is worth our effort given how close the launch deadline is.

Just for your knowledge tho, the "better" way would be to redirect the user back to the detail page for the change request and display a toast or a banner saying "canceled" or "confirmed", and potentially a loading spinning while it's still loading or something, but that can be done later as part of UX improvement efforts when we're not itching to launch

src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <>
+      <PageTitle title={`Change Request #${crId}`} />
+      <PageBlock
+        title={''}
+        headerRight={<></>}
+        body={
+          <div className={'row'}>
+            <div className={'mx-auto mw-50'}>
+              <h2 className={'text-center mb-3'}>{accepted ? 'Accept' : 'Deny'} Change Request</h2>
```
jamescd18, line 48:
```suggestion
              <h2 className={'text-center mb-3'}>{accept ? 'Accept' : 'Deny'} Change Request</h2>
```

src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx
```diff
@@ -0,0 +1,73 @@
...
+              </Form>
+            </div>
+          </div>
+        }
+      />
+    </>
+  );
+};
+
+export default ReviewChangeRequestsView;
```
jamescd18, line 73:
```suggestion
export default ReviewChangeRequestsView;

```

src/components/change-requests/review-change-request/review-change-request.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter, Route } from "react-router-dom";
+import ReviewChangeRequest from "./review-change-request";
+
+const renderComponent = (accepted: boolean) => {
+    return render(
+        <MemoryRouter initialEntries={['change-requests/37']}>
```
jamescd18, line 12: I believe I wrote helper functions to avoid the need for putting a memory router in the test file. Check out this test file for an example: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1121297ac2263c7c55085c8faef3fd2a759f0ed7/src/components/change-requests/change-requests.test.tsx#L41

jamescd18 (changes requested):
Absolutely love this! There's just a few minor things here and there. Really well done with this! Good usage of the mock for testing just the routing, and great test cases for the view component.

review comments:
- rename "accepted" parameter to "accept"
- add newlines to the end of a few files
- remove unused imports
- use the cooler boolean prop shorthand
- remove the memory router (lmk if the example isn't sufficient to understand it, I'm happy to explain how it works)
- add the same router wrapper pattern to the change request details tests so they don't fail

To explain a bit more, I wrote some helpers to make it easier to wrap components under test in a router and then only expose control over the URL that the router is navigated to when it is needed. [This is an example](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1121297ac2263c7c55085c8faef3fd2a759f0ed7/src/components/change-requests/change-requests-table/change-requests-table.test.tsx#L30) where we don't need control over the URL, we just need to wrap the component under test in a router because page navigation `<Link />` components can't be used outside of a router. So basically because you added the action buttons to the change request details page, we now need to go to those tests and wrap the component under test in a router using the helper thingies. This kind of refactor is precisely why it's always good to have the component under test separated out in its own function (the `renderComponent()` functions) in the test file so that it's easier to modify later as needed.

Really good job Justin!

src/components/change-requests/review-change-request/review-change-request/review-change-request.test.tsx
jpwaves: Where would I use `routerWrapperBuilder` in my tests? Most of my tests in this file are for making sure the component renders correctly so I don't really know where I would use this so that I don't get the "declared but never used" error from the linting tools. Is this something I would wrap around my component in the `renderComponent` function?

jamescd18: You want to [use it like this](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1121297ac2263c7c55085c8faef3fd2a759f0ed7/src/components/change-requests/change-request-details/change-request-details.test.tsx#L30) when you don't need to control the navigation URL. You want to [use it like this](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1121297ac2263c7c55085c8faef3fd2a759f0ed7/src/components/change-requests/change-requests.test.tsx#L41) when you do need to control the navigation URL. In either case, it gets used inside of the test file's `renderComponent` helper function.

Can you share a screenshot of the error with the ternary expression? I'm not sure what you mean. We ideally want to stick with importing from `test-utils` and not directly from the package.

src/components/change-requests/change-requests.tsx
jpwaves: Do you think renaming the prop to `option` and having values of `'accept'` or `'deny'` be better and more readable?

src/components/change-requests/review-change-request/review-change-request/review-change-request.test.tsx
jpwaves: Here's the screenshot of the error:
failed-test-cr-review

Here's a screenshot of the line of code that I think is causing the issue:
possible-test-failure-src

src/components/change-requests/change-requests.tsx
jamescd18: Love it!

jamescd18 (approved):
Awesome job! Love to see all the tests passing!

PR #421, opened 2022-01-09
Populates Implemented Changes Section

RChandler234:
I've added the implemented changes list component to solve this issue. The implemented changes section is now populated and I believe the links work properly, I'm not sure if where I added the implemented changes list component is in the wrong spot, and I also need to add testing for it, but I wasn't sure how to do that. I'm working on implementing the final part of the ticket, the tooltip rn.


RChandler234:
image

jamescd18 (commented):
Love this! Great first step on the implementation and great usage of our helper components.

RChandler234:
image
image

jamescd18 (changes requested):
I'd love it if we could have the target for the tool tip be the whole bullet instead of just the link, and then also display the tool tip to the right so it doesn't cover up anything. Other than that it looks great!!

RChandler234:
image
image
The target is now the text instead of the link-I couldn't get it to target the whole thing-and the tooltip displays to the right of the text.

jamescd18 (approved):
LGTM, love it! Really really love the tool tip!!

PR #422, opened 2022-01-10
#318 - added token signing

dli85:


jamescd18:
Okay! I think I found a potential solution. I came across some example code that Netlify provides that suggests that process.env.<VARIABLE NAME> is the proper way to do it and you don’t need a package for it to work in production. So then it’s just local development that is the issue, and we should try this to see if it will fix it: https://github.com/netlify/netlify-lambda/issues/118#issuecomment-506973346

Issue #429, opened 2022-01-11
Work Package - Incorrect data displayed for "Deliverables" section


jpwaves:
### Browser(s)

Firefox

### Platform(s)

Desktop

### Version(s)

main

### Observed Behavior

In a work package, the "Deliverables" section currently displays Description_Bullet data for the "Expected Activities" section.

### Expected Behavior

The "Deliverables" section should display the deliverables for the work package and not the expected activities i.e. the data that should be displayed should come from the `workPackageIdDeliverables` column in the `Description__Bullet` table.

### Steps to Reproduce

1. Go to a work package
2. Access the database via psql through the terminal (this may require additional steps depending on if you have already set up psql access for the user you are using to access the database)
3. Enter SELECT * FROM "Description_Bullet"; into the terminal and then compare the expected detail of a description bullet for a work package deliverable to what is displayed in the page.
Alternatively, you can just look at the "Expected Activities" and "Deliverables" section and see that their displayed data is the same.

### Screenshots (as needed)

![wp-1-incorrect-deliverables](https://user-images.githubusercontent.com/48561685/148880363-781d2bf0-e7df-4e59-bd0f-63ad61594e42.png)
![wp-2-incorrect-deliverables](https://user-images.githubusercontent.com/48561685/148880364-476debf8-9e80-4084-ba25-cb5323976a08.png)
![wp-3-incorrect-deliverables](https://user-images.githubusercontent.com/48561685/148880370-b8da2609-9a93-42b2-9009-24fc6b14cbd6.png)

jamescd18:
Found the offending line of code! Probably a simple copy-paste error. Should be a super quick fix: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/4dd4ef4c52fa736ac2f8b6e9028c247aa7b84d68/src/backend/functions/work-packages.ts#L85

PR #432, opened 2022-01-11
Fix deliverables getting data for expected activities

jpwaves:
Made the change to the copy and paste error noted in https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/429#issuecomment-1009588853. I launched the app and confirmed that they deliverables now correctly appear. Not sure if there's any tests I need to update for this change.


jamescd18 (approved):
LGTM, nice small fix

Issue #431, opened 2022-01-11
Change Request - Need to conditionally show approve/deny buttons only for pending change requests


jpwaves:
### Browser(s)

Firefox

### Platform(s)

Desktop

### Version(s)

main

### Observed Behavior

Approve and deny buttons currently show up for every change request. Having those buttons for an accepted/denied change request doesn't make sense since those change requests have already been either accepted or denied.

### Expected Behavior

The approve and deny buttons should only show up for change requests that haven't been approved or denied yet.

### Steps to Reproduce

1. Go to a change request detail page that has been marked as accepted/not accepted on the change request table.

### Screenshots (as needed)

![cr-1-accepted-but-has-accept-btn](https://user-images.githubusercontent.com/48561685/148881863-1f50ceec-89ad-42c5-92e5-ee1227c3b7f8.png)

PR #433, opened 2022-01-11
#431 - Add condition for when buttons appear

jpwaves:
Added the condition that the approve and deny buttons show up only if the `accepted` column of a change request is not specified. This means that both approved and denied change requests won't have the buttons, but pending change requests will. Currently, the change request table doesn't indicate that a change request is denied, so it could be difficult to tell which change requests have already been reviewed.

Screenshot for approved CR:
accepted-cr

Screenshot for denied CR:
denied-cr

Screenshot for pending CR:
not-a-nor-d


jamescd18 (approved):
LGTM, nice little conditional display

Issue #391, opened 2021-11-23
Change Requests - Split Change Request Detail Component Into Separate Components


kevinyu328:
### Description

The change request details component is currently one large component that houses everything within the page. This large component needs to be split into smaller components to enhance readability and separate concerns. Separating them into different smaller components also helps with testing and future expansion. In the future, we may have more than 3 kinds of type-specific CR details blocks, or we may add additional blocks to the single change request details page, or a number of other potential expansions.

### Background

The blocks on the project and work package details pages are already built as separate components, so you could use those as examples when tackling this issue.

### Proposed Solution

- [ ] Have separate components for each of the blocks that come together to form the single change request details page:
  - [ ] CR details
  - [x] type-specific CR details
  - [x] implemented changes list
- [x] Have separate components for each of the different blocks which could be displayed for the type-specific CR details block:
  - [x] standard CR details
  - [x] activation CR details
  - [x] stage gate CR details

PR #434, opened 2022-01-11
#391 - Separate Change Request Details Component

jpwaves:
I created separate detail components for each type of change request stored in the `type-specific-details` directory and separated the implemented changes list into its own component as well. I mainly copy and pasted the existing code when creating the new components, so there shouldn't be any visual changes. I also copied the tests for each respective component and put them in the appropriate test files; I wasn't sure if this was okay so I currently have the original tests commented out in `change-request-details.test.tsx`. I also changed `buildDetails` to use a switch statement since I thought it was cleaner and would make more sense for the purpose of the function.

Screenshot of an activation CR:
activation

Screenshot of an stage gate CR:
stagegate

Screenshot of an standard CR:
standard


src/components/change-requests/change-request-details/change-request-details/type-specific-details/stage-gate-details/stage-gate-details.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+describe('Change request details stage gate cr display element tests', () => {
+  it('Renders confirm completed', () => {
+    const cr: StageGateChangeRequest = exampleStageGateChangeRequest;
+    renderComponent(cr);
+    expect(screen.getByText(`Confirm WP Completed`)).toBeInTheDocument();
+    expect(screen.getByText(`${cr.confirmDone ? 'YES' : 'NO'}`)).toBeInTheDocument();
+  });
+
+  it('Renders leftover budget', () => {
+    const cr: StageGateChangeRequest = exampleStageGateChangeRequest;
```
jamescd18, line 27: You could penalty just lift this line to be the first line inside the describe so as to eliminate having this line twice

src/components/change-requests/change-request-details/change-request-details/change-request-details.tsx
```diff
@@ -167,7 +87,7 @@ const ChangeRequestDetails: React.FC<ChangeRequestDetailsProps> = ({
         }
       />
       {buildDetails(changeRequest)}
-      <PageBlock title={'Implemented Changes'} headerRight={<></>} body={<>list of changes</>} />
+      <ImplementedChanges />
```
jamescd18, line 90: This change will impact what @RChandler234 is working on, so unless he wants you to make the change then let's not make this change, especially since it a small change.

src/components/change-requests/change-request-details/change-request-details/change-request-details.test.tsx
```diff
@@ -91,81 +80,81 @@ describe('Change request details common display element tests', () => {
...
-  });
-
-  it('Renders leftover budget', () => {
-    const cr: StageGateChangeRequest = exampleStageGateChangeRequest;
-    renderComponent(cr);
-    expect(screen.getByText(`Leftover Budget`)).toBeInTheDocument();
-    expect(screen.getByText(`$${cr.leftoverBudget}`)).toBeInTheDocument();
-  });
-});
+// describe('Change request details standard cr display element tests', () => {
```
jamescd18, line 83: I'd say for all these tests you don't need to remove them all from the parent component, just maybe slim them down a bit. Have at least 1 sanity check test for each type of CR just to ensure the code gets run via a test?

jamescd18 (changes requested):
Overall LGTM since just copy change. Pending review from @RChandler234 and also a couple small items.

src/components/change-requests/change-request-details/change-request-details/change-request-details.test.tsx
jpwaves: Sure, sounds good.

src/components/change-requests/change-request-details/change-request-details/type-specific-details/stage-gate-details/stage-gate-details.test.tsx
jpwaves: I'll look into it and check the other tests to see if there is anything else that can be refactored.

src/components/change-requests/change-request-details/change-request-details/change-request-details.tsx
RChandler234: Yeah, I would appreciate it if we didn't make this change if possible because otherwise I'm going to have to move around all the code I've been working on. If this causes you a bunch of issues let me know, and I can work around it. Otherwise though, the code looks fine/ shouldn't conflict with my PR

jpwaves: Gotcha, I'll revert this change and leave out separating the implemented changes as its own component for this PR and we can just make a ticket to have it separated out in the future potentially.

jamescd18 (approved):
Awesome! LGTM. I really appreciate getting this cleaned up cuz I know it's all my fault for this mess being the way it is (I wrote it lol). This will be so much better moving forward. Thank you Justin!

RChandler234:
Thanks a bunch @jpwaves ! Everything looks awesome, thanks for switching that stuff back, it should be ready to merge now

RChandler234 (commented):
oop, just realized I need to leave my review here. Again everything looks a lot better, thanks for cleaning this stuff up, etc.

Issue #435, opened 2022-01-12
Change Requests - Update Change Requests to Fit Customer Needs


jamescd18:
## Prerequisites
- [x] #419 

## Description

Basically the database has the latest naming for the different types of Change Requests, but the front-end / application types are out of date. Also, we need to make some small modifications to change requests to satisfy customer requirements for MVP launch.

Once the app types enums match the database types enums, then we should be able to remove the converter function and usage from the change requests endpoint. This should work since the backend should just be able to grab the string from the db, stringify it for sending over HTTP, front-end picks it up and `JSON.parse()`, converting it to the app types enum which has the same underlying all caps snake case strings. This will allow us to simplify the back-end.

## Tasks
- [x] rename front-end types to match database
- [x] update test stub data
- [x] ensure enums are working properly in entire app
- [x] remove enum convert from back-end
- [x] Rename `DEFINIITION_CHANGE` to `REDEFINITION`
- [ ] Add `documentationLink` to the Scope Change Request as a `string`
- [ ] Add `DESIGN` as an item in the `Scope_CR_Why_Type` enum

jamescd18:
Discussion with Head of Project Management, Der Chi Lin:

- [ ] Rename `DEFINIITION_CHANGE` to `REDEFINITION`
- [ ] Add `documentationLink` to the Scope Change Request as a `string`
- [ ] Add `DESIGN` as an item in the `Scope_CR_Why_Type` enum

Issue #397, opened 2021-12-21
Change Requests - Create Change Requests


kevinyu328:
### User Stories

As an engineering lead, I want the ability to create change requests in-app, so that I can easily communicate the changes that need to be made to the PMO.

### Background

As the engineers work on the car, things may not go as originally planned, as there may be changes to the original project or work packages that were created during the project definition process. Thus, in order to document and communicate these changes to the PMO for authorization, engineers need to be able to create change requests.

### Tickets

 - [x] #196
 - [x] #192
 - [x] #193
 - [x] #194
 - [x] #195
 - [x] #412
 - [x] #328
 - [x] #435
 - [x] #481

jamescd18:
Done. See #636 for final form

PR #436, opened 2022-01-12
Standardize enums

### LINKED ISSUE #419: General - Standardize Enums
### Desired Changes

An annoying amount of errors come up because enums are created in various different ways throughout the application. It would be so much easier if they were all just standardized as all caps strings or as numbers, depending on the particular need.

### Screenshots (as needed)

_No response_

jamescd18:
PR for my initial work on standardizing the enums. started by doing the standardization, but I haven't done anything more. I realize that I want to rename the front-end enums to match the back-end enums, but I think that's technically a separate ticket #435 which is part of epic #397. closes #419.

### DIFF
--- a/src/components/projects/projects-table/projects-table.test.tsx
+++ b/src/components/projects/projects-table/projects-table.test.tsx
@@ -4,7 +4,7 @@
  */
 
 import { UseQueryResult } from 'react-query';
-import { Project } from 'utils';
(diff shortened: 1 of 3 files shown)


RChandler234 (commented):
Looks like a good, straightforward change to me. I like how you updated the tests as well. I'd say merge it whenever you're ready!

Issue #428, opened 2022-01-11
Change Request - Add Review Notes to the Details Page


jpwaves:
### Desired Changes

Currently, the review notes for a change request aren't visible on the details page, if review notes exist for a change request. We would like to add this so users can see any review notes made for accepted or denied change requests. This will likely result in adding a new component to the change request details page.

IMPORTANT: This ticket should be done once #391 is completed.

PR #437, opened 2022-01-12
#428 - Add review notes to details page

jpwaves:
I added the review notes to be above the "Implemented Changes" list on the CR details page. Currently, it displays the review notes if they exist, and if there are no review notes, it fills the block in with a placeholder text. An alternative to this would be to conditionally show the review notes only if review notes exist for the change request. This could reduce the clutter of blocks on the page for change requests that haven't been reviewed yet. Let me know your thoughts about the alternative.

Has review notes:
has-rn

Doesn't have review notes:
no-rn


jamescd18 (approved):
This is good enough for MVP launch 👍 thank you Justin!

PR #440, opened 2022-01-14
#318 add session jwt validation

dli85:
Northeastern-Electric-Racing/FinishLine#9 added jwt token signing as well as verifying


package.json
```diff
@@ -186,6 +188,7 @@
     "@babel/preset-typescript": "^7.12.13",
     "@testing-library/react-hooks": "^5.1.2",
     "@types/aws-lambda": "^8.10.89",
+    "@types/jsonwebtoken": "^8.5.7",
```
jamescd18, line 191: I think this should be a [dev dependency](https://medium.com/@dylanavery720/npmmmm-1-dev-dependencies-dependencies-8931c2583b0c)

package.json
```diff
@@ -166,7 +166,9 @@
     "@types/react-router-dom": "^5.1.7",
     "axios": "^0.21.1",
     "bootstrap": "^4.6.0",
+    "dotenv": "^10.0.0",
```
jamescd18, line 169: This might also be a dev dependency. The Netlify functions example code suggests it is not needed in production I think  https://www.netlify.com/products/functions/

src/backend/functions/users.ts
```diff
@@ -89,6 +92,8 @@ const logUserIn: ApiRouteFunction = async (_params, event) => {
     user = createdUser;
   }
 
+  const accessToken = jwt.sign(user, process.env.JWT_SECRET as string, { expiresIn: '20mins', algorithm: 'HS256' });
```
jamescd18, line 95: How did you decide on the algorithm and the expiration? What are the options?

src/backend/functions/users.ts
```diff
@@ -97,7 +102,7 @@ const logUserIn: ApiRouteFunction = async (_params, event) => {
     }
   });
 
-  return buildSuccessResponse(usersTransformer(user));
+  return { ...buildSuccessResponse(usersTransformer(user)), accessToken };
```
jamescd18, line 105: Is this the best way to be passing it to the front-end? Did you consider doing a header? I think it has implications on our type interfaces, how the front-end retrieves it, how the front end can send it back, and how this needs to work during local development too. Right now I think the local development login skips sending a POST request to the login endpoint and instead just lets the user into the front end application and fills in the user object on the front end

src/utils/src/api-utils.ts
```diff
@@ -98,4 +100,21 @@ export const routeMatcher = (
...
   return buildNotFoundResponse("route", event.path);
-  };
\ No newline at end of file
+};
+
+/**
+* Verifys a token and returns the payload/user if successful, otherwise returns the error
+* @param token The jwt token
+*/
+export const verifyToken = (
```
jamescd18, line 109: How does the consumer of this function determine whether the output is a success vs an error? I think it might be better to ensure the consumer of the function either doesn't need to figure this out, or it is really easy for them to know

jamescd18 (changes requested):
This is a great baseline and a fantastic first step! I think it would be good to discuss / consider / brainstorm the broader design implications that the specifics of this implementation have on the rest of the application now that you have a baseline working JWT creation and verification.

src/backend/functions/users.ts
dli85: The options are HS256 and RS256. HS256 uses a shared secret while RS256 uses a public and a private key

package.json
acd124: This might even not be necessary at all, since netlify also provides env variables when serving at localhost

package.json
acd124: This might also suggest that other `@types` dependencies that are currently in regular dependencies should be moved to dev dependencies

src/utils/src/api-utils.ts
dli85: I will change the function so it only returns the decoded payload if successful, and nothing if not. Consumers can then test if the return value is truthy it know there was not an error

src/backend/functions/users.ts
acd124: Note: this method of trying to send the token to the client does not appear to work

package.json
jamescd18: yes, but this ticket should just focus on new ones being added

edwinliiiii: @jamescd18 is it a dev dependency if it's under "devDependencies" or is there more to it

jamescd18: Nope that's it!

src/backend/functions/users.ts
jamescd18: Awesome!

src/utils/src/api-utils.ts
jamescd18: that should work

src/backend/functions/users.ts
```diff
@@ -89,6 +92,10 @@ const logUserIn: ApiRouteFunction = async (_params, event) => {
     user = createdUser;
   }
 
+  const accessToken = jwt.sign(user, process.env.JWT_SECRET as string, { expiresIn: '20mins', algorithm: 'HS256' });
+
+  event.headers["accessToken"] = accessToken;
```
jamescd18, line 97: I think this refers to the request event headers (correct me if I'm wrong), but we want to put it into the response headers. Also, why name the header "accessToken"? You're probably going to need to use the `buildResponse` function I wrote. Also, you'll want to research how the front-end is going to pick up the token and consider how that will impact things

jamescd18 (changes requested):
This is looking better! I love to see how this is evolving. I left some review comments, but the main thing now is to start thinking about how the front-end will pick up the token properly and send it back for all the other calls.

package.json
```diff
@@ -197,4 +198,4 @@
     "prisma": "~2.20.0",
     "ts-node": "^9.1.1"
   }
-}
+}
```
jamescd18, line 201:
```suggestion
}

```
always have a blank newline at the end of the file

src/utils/src/api-utils.ts
```diff
@@ -98,4 +100,20 @@ export const routeMatcher = (
...
+  token: any,
+) => {
+  jwt.verify(token, process.env.JWT_SECRET as string, (err: any, decoded: any) => {
+    if (decoded) {
+      return decoded;
+    } else {
+      console.log(err);
+    }
+  });
+}
```
jamescd18, line 119:
```suggestion
}

```
same

jamescd18 (changes requested):
Using axios in the back-end is not the correct thing to do. Axios is what we currently use to initiate HTTP requests from the front-end to the back-end. The back-end responds to these HTTP requests with a response. The back-end never initiates HTTP requests to the front-end, nor does it typically ever intimate HTTP requests to itself. Have you done research into HTTP request and response headers?

src/utils/src/api-utils.ts
```diff
@@ -98,4 +100,20 @@ export const routeMatcher = (
...
+* Verifys a token and returns the payload/user if successful.
+* @param token The jwt token
+*/
+export const verifyToken = (
+  token: any,
+) => {
+  jwt.verify(token, process.env.JWT_SECRET as string, (err: any, decoded: any) => {
+    if (decoded) {
+      return decoded;
+    } else {
```
jamescd18, line 115: potentially throw an error?

jamescd18 (commented):
Perfect! I think this is exactly what we want to see on the back-end. Now to examine how it will get picked up, stored, and sent back to the back-end on the front-end. Lmk if you need any pointers. Keep it up guys!

Issue #400, opened 2021-12-21
Change Requests - Populate Implemented Changes Section  


kevinyu328:
### Description

The "Implemented Changes" section on each change request detail page currently has placeholder text that says "list of changes". This section needs to be populated with the actual list of changes that have already been implemented. 

### Background

This feature will be an unordered list similar to the "Goals" and "Features" section in the project detail page and the "Changes" and "Deliverables" section in the work package details page. Feel free to refer to those components for implementation inspiration.  

### Proposed Solution

- [x] Display the implemented changes for a change request in the format displayed in the mocks
- [x] The WBS number for each implemented change should be a link to the respective work package
- [ ] When hovering over the details of the implemented changes, a tooltip should appear displaying the implementer and date-time that the change was implemented
  - [ ] The tooltip should have a black background with white text

### Mocks
![IMG_5723](https://user-images.githubusercontent.com/59738829/147620043-9aeb38f1-7691-4630-95cc-ede4eada6fa0.jpg)

PR #441, opened 2022-01-14
#400 - set WBS_ELEMENT status default to INACTIVE, reset migration fo…

RChandler234:
Made the default for WBS_ELEMENT status INACTIVE, then recreated the database using the new schema after deleting the old migration folder and database.


jamescd18 (approved):
LGTM!

PR #442, opened 2022-01-14
#372 - Change Request Review Endpoint

### LINKED ISSUE #372: Change Requests - Review Change Request Endpoint
## Description
Create a new endpoint route in the back-end that edits the fields of a change request needed for reviewing change requests. 

## Background
The application does not currently support the action of editing change requests. Editing change requests is needed so that the review notes, review date, and accepted/deny status fields of the change request can be changed when project managers are reviewing change requests. 

## Proposed Solution 
Create a new route in the change request lambda endpoint. This will require a new route function and adding the function to the routes array. Pay attention to nested object creation with Prisma. Also note that some not-specified fields must be set.
- [ ] PATCH endpoint on the path string "/change-requests/edit/{ id }"
- Accepted payload has: 
  - [x] reviewNotes string
  - [x] dateReviewed DateTime
  - [x] accepted boolean 
- Method will do the following:
  - [x] Validate incoming data (ensure all fields are present)
  - [x] Select the work package and edit the associated fields/objects in the database
  - [x] Return the edited work package from the database

acd124:
This creates a change request review endpoint following the structure used in the new change request endpoint, validating the body and updating the change request in the database.

Closes #372.

### DIFF
--- a/src/backend/functions/change-requests-review.ts
+++ b/src/backend/functions/change-requests-review.ts
@@ -16,7 +16,7 @@ const prisma = new PrismaClient();
 // handle reviewing of change requests
 export const reviewChangeRequest: Handler = async ({ body }, _context) => {
   // TODO: validate authorization
-  const { crId, reviewNotes, dateReviewed, accepted } = body;
+  const { crId, reviewNotes, accepted } = body;
 
   // ensure existence of change request
   const foundCR = prisma.change_Request.findUnique({ where: { crId } });
@@ -25,7 +25,7 @@ export const reviewChangeRequest: Handler = async ({ body }, _context) => {
   // update change request
   const update = prisma.change_Request.update({
     where: { crId },
-    data: { reviewNotes, accepted, dateReviewed: new Date(dateReviewed) }
+    data: { reviewNotes, accepted, dateReviewed: new Date() }
   });
 
   // TODO: consider transformer or other possible return body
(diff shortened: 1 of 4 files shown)


src/backend/functions/change-requests-review.ts
```diff
@@ -0,0 +1,58 @@
...
+  const { crId, reviewNotes, dateReviewed, accepted } = body;
+
+  // ensure existence of change request
+  const foundCR = prisma.change_Request.findUnique({ where: { crId } });
+  if (!foundCR) return buildNotFoundResponse('change request', `#${crId}`);
+
+  // update change request
+  const update = prisma.change_Request.update({
+    where: { crId },
+    data: { reviewNotes, accepted, dateReviewed: new Date(dateReviewed) }
```
jamescd18, line 28: Let's not take in `dateReviewed` since fewer user inputs make it better. Let's just assume date reviewed is the immediate time when the code here is being executed. effectively `now()`

src/backend/functions/change-requests-review.ts
```diff
@@ -0,0 +1,58 @@
...
+      properties: {
+        crId: { type: 'number' },
+        reviewNotes: { type: 'string' },
+        dateReviewed: { type: 'string', format: 'date' },
+        accepted: { type: 'boolean' }
+      },
+      required: ['crId', 'reviewNotes', 'dateReviewed', 'accepted']
+    }
+  },
+  required: ['body']
```
jamescd18, line 50: Oops i think i forgot this on mine. Wanna just slip it in and add it?

jamescd18 (changes requested):
Mostly looks perfect. Couple small changes

jamescd18 (changes requested):
This looks great! Could we get some tests?

jamescd18 (approved):
LGTM!

Issue #366, opened 2021-10-17
Projects - Project Create Endpoint


KincentLan:
## Description
Create a new endpoint route in the back-end that creates a new project.

## Background
The database has a table for storing projects and we have an interface for what a project is on the front-end. We now need a way to create new projects.

## Proposed Solution
Create a new route in the projects lambda endpoint. This will require a new route function and adding the function to the routes array. Pay attention to nested object creation with Prisma. Also note that some not-specified fields must be set.

- [ ] POST endpoint on the path string "/projects"
- Accepted payload has: 
  - [x] creator user ID number `userId`
  - [x] project name string `name`
  - [x] car number to associate this project under `carNumber`
  - [x] summary string `summary`
- Method will do the following:
  - [x] Validate incoming data (all fields are present)
  - [x] Create the project and associated objects in the database
    - [x] set `projectNumber` on the created `wbsElement` to be the next integer after the highest existing project number for the associated car. (Ex: car 1 has projects 1, 2, 3 then the new project for car 1 will be project 4)
    - [x] set `workPackageNumber` on the created `wbsElement` to be 0 (always 0 for projects)
  - [x] Return the created project from the database

PR #445, opened 2022-01-15
#366 - Add create project endpoint

jpwaves:
The handler for creating a project was modified to take in `crId` since you create projects from change requests. I added helpers to get the max value of a project number for a specific `carNumber` and check if the specific change request was reviewed. Currently, the tests only ensure that the body of the input is valid and that the associated change request exists and has been reviewed. 

For checking if a change request exists, I currently have the helper pass down `null` to `baseHandler` if it doesn't exist and have the failure response returned there. I'm not sure if this is the preferred way of handling this error catching.

Also, for some reason, I had to specify the status to be `INACTIVE` even though it has a default value set in the prisma schema. Not sure if that's only for me though.

### DIFF
--- a/src/backend/functions/projects-new.ts
+++ b/src/backend/functions/projects-new.ts
@@ -8,7 +8,7 @@ import jsonBodyParser from '@middy/http-json-body-parser';
 import httpErrorHandler from '@middy/http-error-handler';
 import validator from '@middy/validator';
 import { Handler } from 'aws-lambda';
-import { PrismaClient, WBS_Element_Status } from '@prisma/client';
+import { PrismaClient } from '@prisma/client';
 import { buildClientFailureResponse, buildNotFoundResponse, buildSuccessResponse } from 'utils';
 
 const prisma = new PrismaClient();
@@ -49,7 +49,7 @@ export const baseHandler: Handler = async ({ body }, _context) => {
 
   // check if the change request for the project was reviewed
   if (!crReviewed) {
-    return buildClientFailureResponse('Cannot implement an unreview change request');
+    return buildClientFailureResponse('Cannot implement an unreviewed change request');
   }
 
   // create the wbs element for the project and document the implemented change request
@@ -60,7 +60,6 @@ export const baseHandler: Handler = async ({ body }, _context) => {
       projectNumber: maxProjectNumber + 1,
       workPackageNumber: 0,
       name: body.name,
-      status: WBS_Element_Status.INACTIVE,
       project: {
         create: {
           summary: body.summary,
@@ -80,7 +79,15 @@ export const baseHandler: Handler = async ({ body }, _context) => {
     },
   });
 
-  return buildSuccessResponse(createdProject);
+  return buildSuccessResponse(
+    {
+      wbsNumber: {
+        car: createdProject.carNumber,
+        project: createdProject.projectNumber,
+        workPackage: createdProject.workPackageNumber,
+      }
+    }
+  );
 }
 
 const inputSchema = {
--- a/src/backend/functions/projects-new.ts
+++ b/src/backend/functions/projects-new.ts
@@ -9,7 +9,12 @@ import httpErrorHandler from '@middy/http-error-handler';
 import validator from '@middy/validator';
 import { Handler } from 'aws-lambda';
 import { PrismaClient } from '@prisma/client';
-import { buildClientFailureResponse, buildNotFoundResponse, buildSuccessResponse } from 'utils';
+import {
+  buildClientFailureResponse,
+  buildNotFoundResponse,
+  buildSuccessResponse,
+  createProjectPayloadSchema
+} from 'utils';
 
 const prisma = new PrismaClient();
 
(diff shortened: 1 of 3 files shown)


src/backend/functions/projects-new.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  // check if the change request for the project was reviewed
+  if (!crReviewed) {
+    return buildClientFailureResponse('Cannot implement an unreview change request');
+  }
+
+  // create the wbs element for the project and document the implemented change request
+  const maxProjectNumber = await getHighestProjectNumber(body.carNumber);
+  const createdProject = await prisma.wBS_Element.create({
```
jamescd18, line 57: I love how here you're creating the WBS element first and nesting the project inside because of how the other things need to be created

src/backend/functions/projects-new.ts
```diff
@@ -0,0 +1,108 @@
...
+        },
+      },
+    },
+    include: {
+      project: true,
+      changes: true,
+    },
+  });
+
+  return buildSuccessResponse(createdProject);
```
jamescd18, line 83: because we're not going to use the returned data for anything, and instead we're just going to boot people over to the page for the project, let's not return the whole thing, let's just return the WBS number.
```suggestion
  return buildSuccessResponse({ wbsNumber: { ... } });
```

jamescd18 (changes requested):
this is looking great! amazing and nearly complete. there's a few adjustments to make, and I believe the tests are failing right now. You'll want to reduce the tests down to only running tests that are guaranteed to not make it to running prisma code. It will fail because we have not properly gotten a way to mock out prisma for our tests

jpwaves:
> this is looking great! amazing and nearly complete. there's a few adjustments to make, and I believe the tests are failing right now. You'll want to reduce the tests down to only running tests that are guaranteed to not make it to running prisma code. It will fail because we have not properly gotten a way to mock out prisma for our tests

I'll comment out the tests that require prisma for now then.

src/backend/functions/projects-new.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  // create the wbs element for the project and document the implemented change request
+  const maxProjectNumber = await getHighestProjectNumber(body.carNumber);
+  const createdProject = await prisma.wBS_Element.create({
+    data: {
+      carNumber: body.carNumber,
+      projectNumber: maxProjectNumber + 1,
+      workPackageNumber: 0,
+      name: body.name,
+      status: WBS_Element_Status.INACTIVE,
```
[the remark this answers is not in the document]
jpwaves: When I remove defining the status of the WBS Element, VS Code red underlines `data` and states the following:
```
(property) data: (Prisma.Without<Prisma.WBS_ElementCreateInput, Prisma.WBS_ElementUncheckedCreateInput> & Prisma.WBS_ElementUncheckedCreateInput) | (Prisma.Without<...> & Prisma.WBS_ElementCreateInput)
The data needed to create a WBS_Element.

Type '{ carNumber: any; projectNumber: number; workPackageNumber: number; name: any; project: { create: { summary: any; }; }; changes: { create: { changeRequestId: any; implementerId: any; detail: string; }; }; }' is not assignable to type '(Without<WBS_ElementCreateInput, WBS_ElementUncheckedCreateInput> & WBS_ElementUncheckedCreateInput) | (Without<...> & WBS_ElementCreateInput)'.
  Type '{ carNumber: any; projectNumber: number; workPackageNumber: number; name: any; project: { create: { summary: any; }; }; changes: { create: { changeRequestId: any; implementerId: any; detail: string; }; }; }' is not assignable to type 'Without<WBS_ElementUncheckedCreateInput, WBS_ElementCreateInput> & WBS_ElementCreateInput'.
    Property 'status' is missing in type '{ carNumber: any; projectNumber: number; workPackageNumber: number; name: any; project: { create: { summary: any; }; }; changes: { create: { changeRequestId: any; implementerId: any; detail: string; }; }; }' but required in type 'WBS_ElementCreateInput'.ts(2322)
index.d.ts(12757, 5): 'status' is declared here.
index.d.ts(8714, 5): The expected type comes from property 'data' which is declared here on type '{ select?: WBS_ElementSelect | null | undefined; include?: WBS_ElementInclude | null | undefined; data: (Without<WBS_ElementCreateInput, WBS_ElementUncheckedCreateInput> & WBS_ElementUncheckedCreateInput) | (Without<...> & WBS_ElementCreateInput); }'
```
For whatever reason, it wants me to define what `status` is. I think it might be that VS Code isn't updated with the types or something like that. I've already run `npm i` again to rebuild the types I have locally and the issue is still there. Do you have any ideas on why I'm bugged for excluding `status`?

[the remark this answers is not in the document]
jamescd18: Weird! We can just leave it then

src/backend/tests/projects-new.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      });
+
+      it('fails when summary is not a string', async () => {
+        const res = await func({ body: { ...goodProjectBody, summary: {} } });
+        expect(res.statusCode).toBe(400);
+        expect(res.body).toBe('Event object failed validation');
+      });
+    });
+
+    // describe('base handler', () => {
```
jamescd18, line 62: use `it.skip(...)` instead!

src/backend/functions/projects-new.ts
```diff
@@ -0,0 +1,115 @@
...
+        workPackage: createdProject.workPackageNumber,
+      }
+    }
+  );
+}
+
+const inputSchema = {
+  type: 'object',
+  properties: {
+    body: {
```
jamescd18, line 96: move to utils like done in https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/eb6e7201d9f43b142a66efb5a01a658375626a8b/src/backend/functions/change-requests-review.ts#L43

jamescd18 (changes requested):
use `it.skip` and resolve the review comments once they've been acknowledged. also we're now moving most of the payload schema to the utils package

jamescd18 (approved):
LGTM!!

Issue #399, opened 2021-12-21
Projects - Work Package Summary Fields


kevinyu328:
### Description

The work package summary that is displayed when users expand a work package could be populated with more useful data, so that project managers can easily see the details of a work package without navigating to another page.

### Background

This component is already populated with some details of the work package. All that is needed to be done is to edit the arrangement of the details and add some extra details that would be useful to project managers.

### Proposed Solution

Fields that need to be included in the summary:
- [x] Start Date
- [x] Due Date
- [ ] Deliverables
- [ ] Expected Activities
- [x] Dependencies

### Mocks

WIP

PR #461, opened 2022-01-20
#399 - Add deliverables & expected activities to wp summary

jpwaves:
In order to add deliverables and expected activities to the work package summary, I had to revert the several type usages from `WorkPackageSummary` back to `WorkPackage`.  I also had to modify the `workPackage` field on the `Project` type to use `WorkPackage[]` instead of `WorkPackageSummary[]`, which meant I had to do some modifications to the project transformer. I'm not completely sure I did the includes correctly, but it seems to work right now. I'm open to suggestions on how to improve the changes I made to the project transformer.

Note: @jamescd18 and I discussed how many deliverables and expected activities to display in the summary, and we decided on 3 being the max, which you will see reflected in the screenshots.

Screenshots of the new work package summary:
Work packages with 3 or less deliverables/expected activities
normal-and-less

Work package with more than 3 expected activities
limit-to-3

### DIFF
--- a/src/backend/functions/projects.ts
+++ b/src/backend/functions/projects.ts
@@ -28,6 +28,7 @@ import {
   WbsElementStatus,
   DescriptionBullet
 } from 'utils';
+import { changeRequestTransformer } from '../../services/transformers/change-requests.transformers';
 
 const prisma = new PrismaClient();
 
@@ -43,7 +44,14 @@ const manyRelationArgs = Prisma.validator<Prisma.ProjectArgs>()({
     goals: true,
     features: true,
     otherConstraints: true,
-    workPackages: { include: { wbsElement: true, dependencies: true } }
+    workPackages: {
+      include: {
+        wbsElement: { include: { changes: { include: { implementer: true } } } },
+        dependencies: true,
+        expectedActivities: true,
+        deliverables: true
+      }
+    }
   }
 });
 
@@ -54,7 +62,14 @@ const uniqueRelationArgs = Prisma.validator<Prisma.WBS_ElementArgs>()({
         goals: true,
         features: true,
         otherConstraints: true,
-        workPackages: { include: { wbsElement: true, dependencies: true } }
+        workPackages: {
+          include: {
+            wbsElement: { include: { changes: { include: { implementer: true } } } },
+            dependencies: true,
+            expectedActivities: true,
+            deliverables: true,
+          }
+        }
       }
     },
     projectLead: true,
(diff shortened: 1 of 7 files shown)


src/utils/src/types/project-types.ts
```diff
@@ -42,7 +42,7 @@ export interface Project extends WbsElement {
   goals: DescriptionBullet[];
   features: DescriptionBullet[];
   otherConstraints: DescriptionBullet[];
-  workPackages: WorkPackageSummary[];
+  workPackages: WorkPackage[];
 }
 
 export interface WorkPackageSummary {
```
acd124, line 48: This interface has now been refactored out of service, and can be removed, with it also the workPackageSummaryTransformer can be removed. Make sure to check for any leftover references to things you remove.

src/components/projects/wbs-details/project-container/work-package-summary/work-package-summary.test.tsx
```diff
@@ -36,6 +36,12 @@ describe('Rendering Work Package Summary Test', () => {
     expect(
       screen.getByText(`${endDatePipe(wp.startDate, wp.duration)}`, { exact: false })
     ).toBeInTheDocument();
+    wp.expectedActivities.slice(0, 3).map((expectedActivity) => {
```
acd124, line 39: Since we don't care about the return values for these tests, you can just use forEach and the linter will also stop complaining

src/backend/functions/projects.ts
```diff
@@ -64,11 +78,11 @@ const uniqueRelationArgs = Prisma.validator<Prisma.WBS_ElementArgs>()({
 });
 
 const convertStatus = (status: WBS_Element_Status): WbsElementStatus =>
-  ({
```
acd124, line 67: I'm pretty sure the spacing here was previously done automatically by prettier using the settings specified in the package.json, I believe there was also a minor comma thing I noticed that prettier would also remove. This is no issue at all with the code, I just want to point it out.

acd124 (changes requested):
Overall great work with the prisma includes, it does look a little complicated, but that's just the way the data is structured in a somewhat self-referential manner. I'm glad to see the WorkPackageSummary being cleaned out now, just a couple more changes to take care of now unused functions. I do wonder if we want to have an ellipsis or other indicator that there are more than 3 of the expected activities or deliverables, but that's a design question that I don't have the answer to so I will not specifically request that change.

src/utils/src/types/project-types.ts
jpwaves: Sounds good, I'll remove those

src/components/projects/wbs-details/project-container/work-package-summary/work-package-summary.test.tsx
jpwaves: Ah, forEach makes more sense in this case, will change

jpwaves:
> Overall great work with the prisma includes, it does look a little complicated, but that's just the way the data is structured in a somewhat self-referential manner. I'm glad to see the WorkPackageSummary being cleaned out now, just a couple more changes to take care of now unused functions. I do wonder if we want to have an ellipsis or other indicator that there are more than 3 of the expected activities or deliverables, but that's a design question that I don't have the answer to so I will not specifically request that change.

This is something that @jamescd18 and I also discussed. For the current ticket, we decided it wasn't necessary for the initial launch. However, we can open up a ticket in the future to add that ellipsis feature in.

jamescd18 (approved):
LGTM!

Issue #321, opened 2021-08-28
Docs - When to push package.json?


VerySeriousHireableAccount:
When to push package.json?

VerySeriousHireableAccount:
answer: only push it when you make changes to the dev dependencies/dependencies

jamescd18:
Add something to the contributor docs to note this ^

PR #462, opened 2022-01-20
#321 - When to push package.json?

Ari-S-123:
Added the necessary documentation to `Onboarding.md`


jamescd18 (changes requested):
Ooops, my bad, it should actually go in the Contributor Guide, sorry!

Also, for docs, we want to do one sentence one line. This helps when editing in the future.

Also, can you please add a thing that tells people to NOT add and commit the package-lock.json changes unless they're making changes to the package.json?

Issue #486, opened 2022-01-22
Projects - Work Package Creation Fails When Project Has 0 Existing Work Packages


jamescd18:
### Browser(s)

Chrome, Firefox, Safari, Other (specify below)

### Platform(s)

Desktop, iOS, Android

### Version(s)

main, production

### Observed Behavior

API returns a type error when trying to create a work package on the `work-packages-create` endpoint for a project that has 0 existing work packages.

### Expected Behavior

Be able to create a new work package. Set the first work package number to 1 if the project has 0 work packages.

### Steps to Reproduce

1. Have a project with 0 existing work packages
2. Create a work package within said project
3. See error response

### Screenshots (as needed)

Request and response
<img width="1286" alt="Screen Shot 2022-01-21 at 7 20 15 PM" src="https://user-images.githubusercontent.com/51571627/150616112-9c3fb32a-d049-4705-89b2-6f190ddc53d8.png">

Project in-app
![Screen Shot 2022-01-21 at 7 20 46 PM](https://user-images.githubusercontent.com/51571627/150616140-eeda856a-7a5f-4be6-930c-1f04ca5deb97.png)

PR #487, opened 2022-01-22
#486 - work package creation fails if project has no work packages

anthonybernardi:
I think I fixed the bug but I can't run it to check because I'm having issues with my dev environment

### DIFF
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -38,6 +38,7 @@ export const createWorkPackage: Handler = async ({ body }, _context) => {
       goals: true,
       features: true,
       otherConstraints: true,
+      wbsElement: true,
       workPackages: { include: { wbsElement: true, dependencies: true } }
     }
   });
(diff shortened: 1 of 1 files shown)


acd124:
There's also a bug where it's finding the max work package number for all projects with said project number across all cars, so if there's a project 1 on car 1, and a project 1 on car 2, it will try to find the max of both's work package numbers. I think the cleanest solution is to not query the database again and simply use the work packages provided in `project.workPackages` and reduce it to the max workPackageId

src/backend/functions/work-packages-create.ts
```diff
@@ -38,16 +38,20 @@ export const createWorkPackage: Handler = async ({ body }, _context) => {
...
 
   if (project === null) throw new TypeError('Project Id not found!');
 
   // eslint-disable-next-line prefer-destructuring
-  const { carNumber, projectNumber } = project.workPackages[0].wbsElement;
+  const { carNumber, projectNumber } = project.wbsElement;
 
-  const workPackageNumber = (await getHighestWorkPackageNumber(projectNumber)) + 1;
+  const workPackageNumber =
+    project.workPackages === undefined || project.workPackages.length === 0
```
acd124, line 52: If you check the types, you'll see that `project.workPackages` has a type of something. like `Work_Packages & {...}` based on the prisma query, indicating that it cannot be undefined. if it ever is undefined we have prisma to blame for lying to us, so there's no need to check.

anthonybernardi: gotcha, makes sense

src/backend/functions/work-packages-create.ts
```diff
@@ -38,16 +24,22 @@ export const createWorkPackage: Handler = async ({ body }, _context) => {
       goals: true,
       features: true,
       otherConstraints: true,
```
acd124, line 26: Not sure why these other properties are being asked for, I know you didn't touch them, but I believe these 3 can be removed without any effect.

acd124 (dismissed):
The bug looks fixed, and fewer requests are being made to the database. Great work, looks good to me 👍🏼

src/backend/functions/work-packages-create.ts
anthonybernardi: true, good point

jamescd18 (approved):
Much cleaner, I love it!

Issue #368, opened 2021-10-17
Projects - Work Package Edit Endpoint


jamescd18:
## Description
Create a new endpoint route in the back-end that edits a work package. Every single field that is being updated and changed needs to have a change created associated that describes how the field is changing.

## Background
The application does not currently support the action of editing work packages. Currently we have to create a change request that a PM reviews. After they approve, they have to _manually_ edit the fields in the database. We need a way to be able to edit work packages that currently exist.

## Proposed Solution 
Create a new route in the work packages lambda endpoint. This will require a new route function and adding the function to the routes array. Pay attention to nested object creation with Prisma. Also note that some not-specified fields must be set.

@jamescd18 says:
> I imagine that the strategy for getting this to work is to just pull in the incoming fields, grab the existing work package, and then execute a prisma query for each field that changes (since you’ll be creating a new “Change” for each field that is changing) and then just skip executing a query if the field hasn’t changed

- [ ] POST endpoint on the path string "/work-packages-edit"
- Accepted payload has: 
  - [ ] wbsElementId of work package to be edited
  - [ ] userId number of user that edited
  - [ ] name string
  - [ ] crId number
  - [ ] startDate date
  - [ ] duration number
  - [ ] dependencies array of wbsElementIds (can be empty)
  - [ ] expectedActivities array of strings (can be empty)
  - [ ] deliverables array of strings (can be empty)
- Method will do the following:
  - [ ] Validate incoming data (all fields are present)
  - [ ] Select the work package and edit the associated fields/objects in the database
  - [ ] Return the edited work package from the database

### Changes

- `name`, `startDate`, `duration`, `budget`, `projectLead`, `projectManager` fields --> `Changed <name of field> from X to Y.`
- `dependencies` field --> `Added new dependency X.` or `Removed dependency X.`
- `expectedActivities` field or `deliverables` field
  - `Added new <name of field> "X".`
  - `Removed <name of field> "X".`
  - `Edited <name of field> from "X" to "Y".`
-

PR #489, opened 2022-01-23
#368 work package edit endpoint

anthonybernardi:
I added the main functionality and tested inputs only. I haven't done any sort of testing to make sure it works yet.

One issue is that how can I determine if an expected activity or deliverable has been edited since there is no id attached to them? As of now I've left that part out.

### DIFF
--- a/src/backend/functions/work-packages-edit.ts
+++ b/src/backend/functions/work-packages-edit.ts
@@ -20,7 +20,6 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
     userId,
     name,
     crId,
-    projectId,
     startDate,
     duration,
     wbsElementIds,
@@ -52,14 +51,6 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
     'name',
     name,
     originalWorkPackage.wbsElement.name,
-    crId,
-    userId,
-    wbsElementId
-  );
-  const projectIdChangeJson = createChangeJsonNonList(
-    'project id',
-    projectId,
-    originalWorkPackage.projectId,
     crId,
     userId,
     wbsElementId
@@ -111,9 +102,6 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
   if (nameChangeJson !== undefined) {
     changes.push(nameChangeJson);
   }
-  if (projectIdChangeJson !== undefined) {
-    changes.push(projectIdChangeJson);
-  }
   if (startDateChangeJson !== undefined) {
     changes.push(startDateChangeJson);
   }
@@ -128,16 +116,21 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
     .concat(deliverablesChangeJson.changes);
 
   // Update any deleted description bullets to have their date deleted as right now
-  await prisma.description_Bullet.updateMany({
-    where: {
-      descriptionId: {
-        in: expectedActivitiesChangeJson.deletedIds.concat(deliverablesChangeJson.deletedIds)
+  const deletedIds = expectedActivitiesChangeJson.deletedIds.concat(
+    deliverablesChangeJson.deletedIds
+  );
+  if (deletedIds.length > 0) {
+    await prisma.description_Bullet.updateMany({
+      where: {
+        descriptionId: {
+          in: deletedIds
+        }
+      },
+      data: {
+        dateDeleted: new Date()
       }
-    },
-    data: {
-      dateDeleted: new Date()
-    }
-  });
+    });
+  }
 
   // update the work package with the input fields
   const updatedWorkPackage = await prisma.work_Package.update({
@@ -145,7 +138,6 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
       wbsElementId
     },
     data: {
-      projectId,
       startDate,
       duration,
       wbsElement: {
@@ -154,22 +146,29 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
         }
       },
       dependencies: {
+        set: [], // remove all the connections then add all the given ones
         connect: wbsElementIds.map((ele: any) => {
           return { wbsElementId: ele };
-        })
-      },
-      expectedActivities: {
-        create: expectedActivities.map((ele: any) => {
-          return { detail: ele };
-        })
-      },
-      deliverables: {
-        create: deliverables.map((ele: any) => {
-          return { detail: ele };
         })
       }
     }
   });
+
+  addDescriptionBullets(
+    expectedActivitiesChangeJson.addedDetails,
+    updatedWorkPackage.workPackageId,
+    'workPackageIdExpectedActivities'
+  );
+  addDescriptionBullets(
+    deliverablesChangeJson.addedDetails,
+    updatedWorkPackage.workPackageId,
+    'workPackageIdDeliverables'
+  );
+  editDescriptionBullets(
+    expectedActivitiesChangeJson.editedIdsAndDetails.concat(
+      deliverablesChangeJson.editedIdsAndDetails
+    )
+  );
 
   // create the changes in prisma
   await prisma.change.createMany({
@@ -178,6 +177,44 @@ export const editWorkPackage: Handler = async ({ body }, _context) => {
 
   // return the updated work package
   return buildSuccessResponse(updatedWorkPackage);
+};
+
+// helper method to add the given description bullets into the database, linked to the given work package
+const addDescriptionBullets = async (
+  addedDetails: string[],
+  workPackageId: number,
+  descriptionBulletIdField: string
+) => {
+  // add the added bullets
+  if (addedDetails.length > 0) {
+    await prisma.description_Bullet.createMany({
+      data: addedDetails.map((element) => {
+        return {
+          detail: element,
+          [descriptionBulletIdField]: workPackageId
+        };
+      })
+    });
+  }
+};
+
+// edit descrption bullets in the db for each id and detail pair
+const editDescriptionBullets = async (editedIdsAndDetails: { id: number; detail: string }[]) => {
+  // edit the edited bullets if there are any to update
+  if (editedIdsAndDetails.length > 0) {
+    await prisma.$transaction(
+      editedIdsAndDetails.map((element) =>
+        prisma.description_Bullet.update({
+          where: {
+            descriptionId: element.id
+          },
+          data: {
+            detail: element.detail
+          }
+        })
+      )
+    );
+  }
 };
 
 // create a change json if the old and new value are different, otherwise return undefined
@@ -192,7 +229,6 @@ export const createChangeJsonNonList = (
   if (oldValue !== newValue) {
     return {
       changeRequestId: crId,
-      dateImplemented: new Date(),
       implementerId,
       wbsElementId,
       detail: `Edited ${nameOfField} from "${oldValue}" to "${newValue}"`
@@ -230,7 +266,6 @@ export const createListChangesJson = <T>(
   return changes.map((element) => {
     return {
       changeRequestId: crId,
-      dateImplemented: new Date(),
       implementerId,
       wbsElementId,
       detail: `${element.type} ${nameOfField} "${element.element}"`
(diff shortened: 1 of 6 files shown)


jamescd18 (commented):
I'm not sure what you mean when you say that expected activities / deliverables don't have a unique ID. The Prisma schema describes how each individual one has an id and they are never deleted, only updated with a deleted date

src/utils/src/types/work-package-types.ts
```diff
@@ -49,4 +49,52 @@ export const workPackageInputSchemaBody = {
...
+export const workPackageEditInputSchemaBody = {
+  type: 'object',
+  properties: {
+    wbsElementId: { type: 'integer', minimum: 0 },
+    userId: { type: 'integer', minimum: 0 },
+    name: { type: 'string' },
+    crId: { type: 'integer', minimum: 0 },
+    projectId: { type: 'number', minimum: 0 },
+    startDate: { type: 'string', format: 'date' },
+    duration: { type: 'number', minimum: 0 },
```
acd124, line 63: Make sure to use integer if you don't want to allow decimals, Check the database schema to make sure you're specifying the right type.

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,251 @@
...
+    deliverables
+  } = body;
+
+  // get the original work package so we can compare things
+  const originalWorkPackage = await prisma.work_Package.findUnique({
+    where: {
+      wbsElementId
+    },
+    include: {
+      wbsElement: {},
```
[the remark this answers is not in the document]
anthonybernardi: ah yeah I forgot how to do that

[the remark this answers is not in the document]
jamescd18: was this fixed?

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,251 @@
...
+      implementerId,
+      wbsElementId,
+      detail: `Changed ${nameOfField} from ${oldValue} to ${newValue}`
+    };
+  }
+  return undefined;
+};
+
+// create a change json list for a given list (dependencies). Only works if the elements can only be added or deleted
+const createListChangesJson = <T>(
```
jamescd18, line 195: these helper methods can be exported and tested

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,251 @@
...
+  });
+
+  newArray.forEach((element: T) => {
+    seenNew.add(element);
+  });
+
+  const changes: { element: T; type: string }[] = [];
+
+  oldArray.forEach((element: T) => {
+    if (seenNew.has(element)) {
```
jamescd18, line 216: does this mean: for each element in the original array, if it also exists in the new array, then create a change for removing it? Cuz that doesn't sound right to me. I'm a tad confused as to how the code works

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,251 @@
...
+// create a change json list for a given list (dependencies). Only works if the elements can only be added or deleted
+const createListChangesJson = <T>(
+  oldArray: T[],
+  newArray: T[],
+  crId: number,
+  implementerId: number,
+  wbsElementId: number,
+  nameOfField: string
+) => {
+  const seenOld = new Set<T>();
```
jamescd18, line 203: I'm a tad confused how this code works and what the need for converting to a set is

jamescd18 (changes requested):
Good start! I'm a tad confused on how the create changes for array works, and I think we discussed adding the description bullet IDs into the input, plus using the date deleted for deleting bullets. Lmk when you're ready for another review

src/backend/functions/work-packages-edit.ts
anthonybernardi: yeah this was a bug it's actually if !seenNew.has

src/backend/functions/work-packages-edit.ts
anthonybernardi: oh cool thanks

anthonybernardi:
Here are some pictures of the query and then how it turns out in the table:

edit query:
edit query

expected activities table: (note that deleted one has a time of deletion, the edited detail says "edited!!!", and the new one was added)
expected activities

all the changes at the bottom were added with wbsElementId 7
changes

the work package was updated
work package

wbs name was changed
wbs element

acd124:
This wasn't mentioned in the ticket, but I looked through the times and found there are a couple other fields we'll want to be able to edit on a work package, being WBS Element Status, project lead, project manager, order in project, and progress. project lead and project manager both being optional, so there should be a way to leave them as not set.

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,369 @@
...
+    originalWorkPackage.wbsElement.name,
+    name,
+    crId,
+    userId,
+    wbsElementId
+  );
+  const startDateChangeJson = createChangeJsonNonList(
+    'start date',
+    originalWorkPackage.startDate,
+    startDate,
```
acd124, line 61: The json input received will give you the date as a string, and date objects are objects, so the current comparison will always add a change, and comparing date object equality can cause issues since they're objects. You may want to consider checking a formatted string or another representation of these dates.

acd124 (changes requested):
The implementation looks solid, just making sure we don't forget things now.

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,369 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import middy from '@middy/core';
+import jsonBodyParser from '@middy/http-json-body-parser';
+import httpErrorHandler from '@middy/http-error-handler';
+import validator from '@middy/validator';
+import { descBulletConverter } from './projects';
```
jamescd18, line 10: I don't believe this will actually work, but I can test it out to find out if you'd like. One of the tough parts about using a serverless backend is the need for code isolation. The way serverless works is essentially that you take an isolated bundle of code and make a random computer run it. If relative path imports were allowed / worked, then that isolated bundle would have to include all the files that the main file references. This would quickly break down. This is why we built the `utils` package to share code between the backend lambda endpoints because the serverless bundling will pull in package imports. Thus, everything in the `utils` package is included in every backend code bundle. That said, we can't put code that relies on Prisma into the `utils` package because the `utils` package has its own package.json. TL;DR you have to copy paste and duplicate the code, you can't export / import it

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,369 @@
...
+import { buildSuccessResponse, DescriptionBullet, workPackageEditInputSchemaBody } from 'utils';
+
+const prisma = new PrismaClient();
+
+export const editWorkPackage: Handler = async ({ body }, _context) => {
+  const {
+    wbsElementId,
+    userId,
+    name,
+    crId,
```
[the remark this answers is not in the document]
jamescd18: Fail fast and error out if the CR doesn't exist or if it is not an approved CR

src/backend/functions/work-packages-edit.ts
jamescd18: The date Object doesn't appear to have a `.equals()` method, but you can convert a date to "milliseconds since Jan 1, 1970 which can then be subtracted. I don't know if this is the best way to do it though. String comparison should work provided the strings are the same format outputs from a date Object.

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,369 @@
...
+      })
+    });
+  }
+};
+
+// edit descrption bullets in the db for each id and detail pair
+const editDescriptionBullets = async (editedIdsAndDetails: { id: number; detail: string }[]) => {
+  // edit the edited bullets if there are any to update
+  if (editedIdsAndDetails.length > 0) {
+    await prisma.$transaction(
```
jamescd18, line 205: Just curious what the need for this is

src/utils/src/types/work-package-types.ts
jamescd18: Did this get resolved? Duration is intended to be an integer

jamescd18 (changes requested):
Looking good so far. I recall mentioning two display items for changes: WBS numbers instead of WBS Element ID integers, plus truncate times off of displayed dates. 

Echoed what Alexander said and added a couple more comments

src/backend/functions/work-packages-edit.ts
anthonybernardi: It's kind of not it was just in an example I found online and stole lol


Here's what it does: "A transaction is a unit of work that only succeeds if every operation within that transaction succeeds. For example, if you update 200 users within a transaction, each update must succeed - if not, all changes are rolled back and the transaction fails as a whole." which could be useful in general although I'm not sure this scenario needs it

src/backend/functions/work-packages-edit.ts
anthonybernardi: oh huh that's interesting, ok I'll just bring it to this file then

src/utils/src/types/work-package-types.ts
anthonybernardi: oh for some reason I thought a duration could be a decimal woops

src/backend/functions/work-packages-edit.ts
jamescd18: yeah I don't think this scenario needs a transaction, but it's fuckin awesome to see we can do it in prisma cuz I never bothered to look into it. it'll be interesting to explore and see where that might eventually be needed.

anthonybernardi: agreed, I think there could definitely be some good use cases for it

src/backend/functions/work-packages-edit.ts
anthonybernardi: do we care about just the date or the time too?

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,494 @@
...
+  if (progressChangeJson !== undefined) {
+    changes.push(progressChangeJson);
+  }
+  if (wbsElementStatusChangeJson !== undefined) {
+    changes.push(wbsElementStatusChangeJson);
+  }
+
+  if (body.hasOwnProperty('projectManager')) {
+    const projectManagerChangeJson = createChangeJsonNonList(
+      'project lead',
```
jamescd18, line 158: Typo? "Project manager"?

jamescd18 (changes requested):
I think it's just one small typo and then it's ready to go. This is absolutely amazing to see! A real beast of a ticket you've tackled here. Thank you Anthony!!

jamescd18 (dismissed):
This is so awesome to see finalized! Thank you!!

src/backend/functions/work-packages-edit.ts
```diff
@@ -0,0 +1,494 @@
...
+      })
+    });
+  }
+};
+
+// edit descrption bullets in the db for each id and detail pair
+const editDescriptionBullets = async (editedIdsAndDetails: { id: number; detail: string }[]) => {
+  // edit the edited bullets if there are any to update
+  if (editedIdsAndDetails.length > 0) {
+    editedIdsAndDetails.forEach((element) =>
```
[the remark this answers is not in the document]
acd124: This may or may not have been part of what the prisma transaction thing did

acd124 (changes requested):
Mainly making sure things get awaited properly so it won't fail after a response is returned

src/backend/functions/work-packages-edit.ts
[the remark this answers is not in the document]
anthonybernardi: yeah I think that is what it does. it might be overkill but I think it guarantees they all get completed

jamescd18 (approved):
LGTM! Tysm for all the hard work that went into this. My bad on marking this a medium when it's really an XL

Issue #443, opened 2022-01-15
Project - Status Setting of Project Filter not Working


RChandler234:
### Browser(s)

Chrome

### Platform(s)

Desktop

### Version(s)

production

### Observed Behavior

I set a status requirement in the project filter, and no projects showed up on the table.

### Expected Behavior

The table should show all the projects that have that status.

### Steps to Reproduce

1. Navigate to the project page
2. Set the status filter to something other than null
3. Apply the filter

### Screenshots (as needed)

![image](https://user-images.githubusercontent.com/29521172/149605530-3416dbc7-3cf5-4b28-89fa-bcfa9a5d9a1d.png)

PR #490, opened 2022-01-24
#443 - Status Setting Bug

Ari-S-123:
Made the status strings uppercase when they are being compared to the enums during filtering.

### DIFF
--- a/src/components/projects/projects-table/projects-table-filter/projects-table-filter.test.tsx
+++ b/src/components/projects/projects-table/projects-table-filter/projects-table-filter.test.tsx
@@ -11,6 +11,7 @@ import {
   exampleProjectLeadUser,
   exampleProjectManagerUser
 } from '../../../../test-support/test-data/users.stub';
+import { WbsElementStatus } from 'utils';
 
 let temp: any[] = [];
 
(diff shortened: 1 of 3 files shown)


jamescd18 (changes requested):
Although I do think this would probably fix the issue, I think the better way to do it would be to make use of the enum which already have values that are uppercase. Try `Object.values()` if you're trying to pull a list from an enum. https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/66f52cb5b14308586f48af514ce5833142e56ef7/src/utils/src/types/project-types.ts#L27

src/components/projects/projects-table/projects-table-filter/projects-table-filter.tsx
```diff
@@ -8,6 +8,7 @@ import { Button, Card, Dropdown, Form } from 'react-bootstrap';
 import styles from './projects-table-filter.module.css';
 import { User } from 'utils';
 import { fullNamePipe } from '../../../../shared/pipes';
+import { WbsElementStatus } from '../../../../utils/src';
```
jamescd18, line 11: this should be added to the existing `utils` import, not a new import because using relative path imports for the `utils` package is not the proper way to access things inside the `utils` package.

src/components/projects/projects-table/projects-table-filter/projects-table-filter.test.tsx
```diff
@@ -64,16 +64,16 @@ describe('projects table filter component', () => {
   it('checking if data in the status dropdown menu is correct', async () => {
     renderComponent();
     expect(screen.queryByText('None')).not.toBeInTheDocument();
-    expect(screen.queryByText('Active')).not.toBeInTheDocument();
-    expect(screen.queryByText('Inactive')).not.toBeInTheDocument();
-    expect(screen.queryByText('Complete')).not.toBeInTheDocument();
+    expect(screen.queryByText('ACTIVE')).not.toBeInTheDocument();
```
jamescd18, line 67: to make these more flexible, you should be using `WbsElementStatus.Active` and the associated others for all these edited lines.

jamescd18 (changes requested):
Much better approach to fixing the bug. Last suggestion is to make the code more flexible and clean up the imports.

jamescd18 (approved):
Love it! Perfect fix and added cleanliness!

PR #491, opened 2022-01-25
#435 - updated front-end enum names

RChandler234:
I changed the names of the enums and updated all the relevant files and the app works fine with just those changes, but when I removed the converter I got errors on "type" and "why", I'm a little confused by why type is error, but I have no idea why "why" is erroring. Let me know if you guys have any thoughts on how to fix it, I can't figure it out rn, but this ticket should be pretty much done once these two errors are addressed. I'll try to comment on the code I'm talking about below.


src/backend/functions/change-requests.ts
```diff
@@ -69,7 +60,7 @@ const changeRequestTransformer = (
   };
   return {
     ...changeRequest,
-    type: convertChangeRequestType(changeRequest.type),
+    type: changeRequest.type,
```
RChandler234, line 63: This line is erroring saying: 
image

RChandler234:
It won't let me direct comment on "why", but just below type, I'm getting this error on why:
image
Not sure what's causing it, it only came up when I deleted the converter and then removed all its usages

src/backend/functions/change-requests.ts
jamescd18: Okay ew. For whatever reason this (the way Prisma implements enums
```ts
const Name = "OPTION_A" | "OPTION_C" | "OPTION_C"
```
is not the same as this (the way we've written our enums)
```ts
enum Name {
  OptionA = 'OPTION_A',
  OptionB = 'OPTION_B',
  OptionC = 'OPTION_C'
}
```

Perhaps we need to keep the converter, but with the renaming and standardization, hopefully it can be closer to a one-liner. @acd124 would probably have a much better idea of what it could look like to just look up the value from one enum in the other enum.

jamescd18 (changes requested):
Looks great to me. Left comment. Not sure what you mean by "why". If you're unable to comment on a specific line of code in the PR, I'd suggest linking to the line in your branch or in main. You can do "copy permalink" to link directly to a line of code.

src/backend/functions/change-requests.ts
acd124: https://github.com/microsoft/TypeScript/issues/17690 Basically the type system won't allow us to hand wave it through as it is. I think the cleanest solution might involve changing the ChangeRequestType, maybe involving the prisma type in some way.

jamescd18: Okay yeah that's way too much work for such a small thing. Let's try replacing the enum we wrote in `utils` with string literal implementation because I don't want to stir up brining Prisma into the `utils` package

jamescd18: Enums give us the flexibility of opaque values, but then we run into issues because Prisma has it implemented as string literal unions

jamescd18 (changes requested):
Sigh, real design trade offs debating here

src/backend/functions/change-requests.ts
acd124: > Can we just do a reverse lookup? `Object.values...` this one or the next comment https://github.com/microsoft/TypeScript/issues/17690#issuecomment-351489397

I think the issue is that no matter how we manipulate the value given from prisma short of forcing ts to see it as a member of the enum it will still be a string literal. I think if you want to keep a sort of key value enum shape you might turn it into an object as const and say the enum type is `enumObj[string]` which should create the string union properly, then have values be accessed with `enumObj.name`

jamescd18: @acd124 do you want to give it a shot? We can't have Prisma code in the `utils` package and I want to preserve the flexibility that proper enums provide (not having tons of hard-coded strings b/c it being a string literal union). I just wanna see what you can come up with fitting those two requirements.

jamescd18 (approved):
LGTM!

PR #492, opened 2022-01-26
Update Onboarding.md

jonathan-chen10:
Update instructions, found at [the help page](https://code.visualstudio.com/docs/getstarted/settings), to get to settings.json in VS Code


jamescd18 (changes requested):
This is a great start! Typically for setup instructions you can't know for sure what version people will have, so I would recommend keeping both methods instead of replacing the old method with the new method. That way it's a "try A, if A doesn't work, try B" kinda thing. Also, don't forget to tag your commit messages with the ticket number! Instructions are described in the contributor guide.

jonathan-chen10:
Will redo this PR, with changes, for ticket #516 .

PR #493, opened 2022-01-27
#435-updated-scope-change-request-and-added-design-item

rishavsarma5:
Here are screenshots of the changes implemented.
change1

image

Is there any more that I need to have do?


src/components/change-requests/new-change-request/new-change-request-page/standard-form-fields/standard-form-fields.tsx
```diff
@@ -47,22 +47,25 @@ const StandardFormFields: React.FC = () => {
...
-            {['Estimation Error', 'School Work', 'Manufacturing Issues', 'Rules Compliance', 'Other Project/Work Package', 'Other'].map((type) => (
+            {[
+              'Estimation Error',
+              'School Work',
+              'Manufacturing Issues',
+              'Design Issues',
+              'Rules Compliance',
+              'Other Project/Work Package',
+              'Other'
+            ].map((type) => (
```
jamescd18, line 58: this should be driven off of the underlying enum using `Object.values()`

jamescd18 (changes requested):
Adding design is not so simple. there's enums and prisma schema that need to be updated to fully propagate this across the whole application

src/components/change-requests/new-change-request/new-change-request-page/standard-form-fields/standard-form-fields.tsx
```diff
@@ -7,6 +7,7 @@ import { Row, Col, Form, InputGroup, FormControl } from 'react-bootstrap';
 import { exampleAllWorkPackages } from '../../../../../test-support/test-data/work-packages.stub';
 import { wbsPipe } from '../../../../../shared/pipes';
 import styles from './standard-form-fields.module.css';
+import { ChangeRequestReason } from 'utils/lib/types/change-request-types';
```
jamescd18, line 10:
```suggestion
import { ChangeRequestReason } from 'utils';
```

things from utils should only ever be imported with `'utils'`

jamescd18 (changes requested):
There's more than needs to be changed because this one single form checkbox on the front-end impacts everything through the entire stack to the database:

- [ ] [the enum in utils](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/aeb832841b5e1035d88de459bda5eeb804226b09/src/utils/src/types/change-request-types.ts#L57)
- [ ] [the enum in prisma schema](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/aeb832841b5e1035d88de459bda5eeb804226b09/src/backend/prisma/schema.prisma#L35)
- [ ] maybe more I'm forgetting

rishavsarma5:
Im not completely sure what happened with the second commit

jamescd18 (approved):
LGTM

Issue #488, opened 2022-01-23
General - Use json-schema-to-ts to type validated endpoint bodies


acd124:
### Description

The body argument of backend functions created with middy is not currently typed, leaving it as type any. We know what type to expect, so we should type the argument so it's easier to understand and harder to mess up in development.

### Background

With the addition of Middy to the backend structures, we have easy validation of complex request bodies. We also added the `json-schema-to-ts` to help figure out what the types are for the frontend to make sure the right data is being sent, however in the backend we're still just typing the body received as any. We have the ability to generate types from these json schemas we have, so we just need to use them.

### Proposed Solution

Standardize and abstract the methods of creating the backend input schemas for middy. Then use the schemas to provide a stronger type for the body argument of each handler.

Issue #454, opened 2022-01-16
General - Refactor Out Required Body Schema


jamescd18:
### Desired Changes

Right now, every back-end endpoint that takes in data is taking the payload schema from the `utils` package and wrapping it in the same "require the body parameter in the event object" validation schema. Let's refactor this out and make a function in the `utils` package api-utils to wrap whatever schema is provided in the `requireBody` schema.

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/eb6e7201d9f43b142a66efb5a01a658375626a8b/src/backend/functions/change-requests-review.ts#L43

PR #505, opened 2022-02-03
#488 - Add Middy Schema Typings

acd124:
This start adds support functions for schemas with middy lambdas. The goal is to add the typings to all the backend middy functions.

### DIFF
--- a/src/backend/functions/change-requests-new.ts
+++ b/src/backend/functions/change-requests-new.ts
@@ -160,15 +160,7 @@ const standardSchema = bodySchema({
   why: arrayType(
     bodySchema({
       explain: stringType,
-      type: enumType(
-        Scope_CR_Why_Type.ESTIMATION,
-        Scope_CR_Why_Type.MANUFACTURING,
-        Scope_CR_Why_Type.OTHER,
-        Scope_CR_Why_Type.OTHER_PROJECT,
-        Scope_CR_Why_Type.RULES,
-        Scope_CR_Why_Type.DESIGN,
-        Scope_CR_Why_Type.SCHOOL
-      )
+      type: enumType(...Object.values(Scope_CR_Why_Type))
     }),
     1 // min 1 item
   )
--- a/src/backend/functions/change-requests-new.ts
+++ b/src/backend/functions/change-requests-new.ts
@@ -9,7 +9,19 @@ import httpErrorHandler from '@middy/http-error-handler';
 import validator from '@middy/validator';
 import { Handler } from 'aws-lambda';
 import { PrismaClient, CR_Type, Scope_CR_Why_Type } from '@prisma/client';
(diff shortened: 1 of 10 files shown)


jamescd18 (commented):
I like what I'm seeing so far. Takes a bit to wrap my head around what it's doing, but it makes sense. I think it would be useful to add some comments

jamescd18 (commented):
Looking good to me so far! No issues when I pulled and ran in local dev, so I think if you think it's good, go ahead and implement for all endpoints

jamescd18:
Check if this PR also covers #454 pls

acd124:
> Check if this PR also covers #454 pls

Yes, it does, also requiring that requests to middy endpoints contain a user agent string, and in future the `eventSchema` helper function could also require a cookie or other header.

src/backend/functions/change-requests-new.ts
```diff
@@ -105,105 +117,88 @@ const createStageGateChangeRequest = async (
...
+  submitterId: intType,
+  wbsElementId: intType,
+  what: stringType,
+  scopeImpact: stringType,
+  timelineImpact: intType,
+  budgetImpact: intType,
+  why: arrayType(
+    bodySchema({
+      explain: stringType,
+      type: enumType(
```
jamescd18, line 163: is there any way we can have this be simplified to `Object.values(Scope_CR_Why_Type)` or is that too much work? the spread operator might help? kinda just guessing and wondering here

src/utils/src/types/work-package-types.ts
```diff
@@ -4,112 +4,39 @@
...
+    wbsElementId: intType,
+    userId: intType,
+    name: stringType,
+    crId: intType,
+    startDate: dateType,
+    duration: intType,
+    dependencies: arrayType(intType),
+    expectedActivities: arrayType(bodySchema({ id: intType, detail: stringType })),
+    deliverables: arrayType(bodySchema({ id: intType, detail: stringType })),
+    wbsElementStatus: enumType('INACTIVE', 'ACTIVE', 'COMPLETE'),
```
jamescd18, line 34: can this use the proper enum instead of hard-coded strings?

jamescd18 (changes requested):
Looks really good! Just a couple minor things

src/backend/functions/change-requests-new.ts
acd124: Yup, that works great since prisma has enums as a union type and object with key value pairs.

src/utils/src/types/work-package-types.ts
acd124: Yup, if the enum is turned into a union type + map of key values, it could also use just ...Object.values(), but that's for another time

jamescd18 (approved):
LGTM!

Issue #495, opened 2022-01-28
Projects - Form Validation for Create New Project Form


kevinyu328:
### Description

Form validation for create new project form. 

### Background

Associated with #465

### Proposed Solution

To be filled in

Issue #465, opened 2022-01-21
Projects - Create New Project Form


jamescd18:
## Description
Create a front end form for creating new projects. This form should be able to ingest all required data when it is completed and send to [this endpoint](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/src/backend/functions/projects-new.ts). This form should only be accessible from an accepted CR. You only need to make the front end componenets; another ticket will handle collecting and transmitting the data to the endpoint.

## Background
Creating the form needed to create new projects will let users have a spot to continue to keep new projects kept in the dashboard. This form is needed because project managers may end up needing to create a new project when implementing a change rquest

## Proposed solution
Ensure that every form detail is a necessary piece of information for a new project. Create the necessary components for this form, making sure correct input entry is followed. The following need to be included:
 - [ ] Save/Cancel buttons
 - [ ] Project Name
 - [ ] Summary
 - [ ] Car Number for the project
 - [ ] Should only be accessible from a Change Request
 - [ ] Change Request ID

## Mockup
<img src="https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/files/7987345/Scan.Feb.1.2022.pdf">

PR #507, opened 2022-02-04
#495 Create project form validation

jpwaves:
A majority of the validation necessary for the form was implemented in #465. I added extra validation on the inputs to limit the range of acceptable numbers and added template feedback elements for when the form fails (though the form never reaches this because the default form validation catches the errors before it can reach this point). 

A big drawback I found while working on this ticket is that a lot of the in-depth validation I would want to do (e.g. making sure the change request ID actually exists, the car number exists, etc.) would require an external library so we don't make the component messy with validation logic. The react bootstrap docs uses [Formik](https://formik.org/), which seems to be fairly popular, and there is also [react-hook-form](https://react-hook-form.com/) if we want to go down the path of using custom hooks for validation. I think looking into one of those (or some other validation library) would be worth looking into post-mvp.

Screenshots:
example-valildation

### DIFF
--- a/src/components/projects/create-project-form/create-project-form.tsx
+++ b/src/components/projects/create-project-form/create-project-form.tsx
@@ -23,23 +23,26 @@ const CreateProjectForm: React.FC = () => {
               <Row>
                 <Col>
                   <Row>
-                    <Form.Group as={Col} aria-required>
+                    <Form.Group as={Col} controlId='project-name-group' aria-required>
                       <Form.Label htmlFor='project-name'>Project Name</Form.Label>
                       <Form.Control id='project-name' type='text' placeholder='Enter project name...' required />
+                      <Form.Control.Feedback type='invalid'>Please provide a project name.</Form.Control.Feedback>
                     </Form.Group>
                   </Row>
                   <Row>
-                    <Form.Group as={Col} aria-required>
+                    <Form.Group as={Col} controlId='car-number-group' aria-required>
                       <Form.Label htmlFor='car-number'>Car Number</Form.Label>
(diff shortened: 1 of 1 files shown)


src/components/projects/create-project-form/create-project-form.tsx
```diff
@@ -23,23 +23,26 @@ const CreateProjectForm: React.FC = () => {
...
                       <Form.Control id='project-name' type='text' placeholder='Enter project name...' required />
+                      <Form.Control.Feedback type='invalid'>Please provide a project name.</Form.Control.Feedback>
                     </Form.Group>
                   </Row>
                   <Row>
-                    <Form.Group as={Col} aria-required>
+                    <Form.Group as={Col} controlId='car-number-group' aria-required>
                       <Form.Label htmlFor='car-number'>Car Number</Form.Label>
-                      <Form.Control id='car-number' type='text' placeholder='Enter car number...' required />
+                      <Form.Control id='car-number' type='number' min={1} placeholder='Enter car number...' required />
```
jamescd18, line 35: Let's set this min to 0 pls. 0 allows for projects unattached to a car. Something i wanna give PMO the flexibility to decide later how to use. Potentially big business projects we want to experiment tracking, or other things.

jamescd18 (changes requested):
Love this! One quick change and it should be good. I personally think for the app, it's much simpler and easier to not do form validation that requires talking to the back-end at user-input time. The idea being that we're going to catch those errors when the user submits the form anyways, so why bother catching it earlier. Nonetheless, it definitely is an enhancement we can investigate

jamescd18 (approved):
LGTM!

Issue #496, opened 2022-01-28
Projects - Form Validation for Create New Work Package Form 


kevinyu328:
## Description

When creating a new work package, project managers will have to fill out the new work package form. We need to implement front-end validation to ensure that the user inputs and submits proper data. Having front-end validation helps ensure that the information put into the new work package form is correctly formatted.

## Background

Including this in the new work package form will help ensure that the submitted data in the form is consistent with the formatting in the database. This is mainly for project managers who need to create new work packages when implementing a change request. 

### Work Package Details
- [x] Work Package Form field should contain text
- [x] "Project W.B.S # " - element passes the `isProject()` function provided by the `utils` package
- [x] Start Date field should be a real date
- [x] Duration field should contain only a number

### Dependencies
- [x] All elements pass the `validateWBS()` function provided by the `utils` package

### Expected Activites
- [x] No validation needed 

### Deliverables
- [x] No validation needed

PR #522, opened 2022-02-08
#496 Create Work Package Form Validation

jpwaves:
Start date validation is inherent by using a `type='date'` input. `validate` returns `true` if the project WBS # is valid and false if not. I put an input restriction on `duration`'s input to have a minimum possible value of 0. Beyond this, I'm not really sure how else to do the validation since we have no established pattern for doing form validation, so I feel this PR is only partially complete because I anticipate there being several changes made to the way I implemented the form validation.

### DIFF
--- a/src/components/projects/create-wp-form/create-wp-form.tsx
+++ b/src/components/projects/create-wp-form/create-wp-form.tsx
@@ -8,14 +8,6 @@ import { useHistory } from "react-router-dom";
 import { isProject, validateWBS, WbsNumber } from "utils";
 import CreateWPFormView from "./create-wp-form/create-wp-form";
 
-const validate = (e: any) => {
-  // start date should already be a valid date by using a date input
-  console.log(e.target.startDate.value);
-
-
-  return true;
-};
-
 export interface EditableTextInputListUtils {
   add: (val: any) => void;
   remove: (idx: number) => void;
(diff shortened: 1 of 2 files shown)


src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -5,8 +5,28 @@
 
 import { useState } from "react";
 import { useHistory } from "react-router-dom";
+import { isProject, validateWBS, WbsNumber } from "utils";
 import CreateWPFormView from "./create-wp-form/create-wp-form";
 
+const validate = (e: any) => {
```
jamescd18, line 11: I think we can simplify things and put this logic into the handle submit function. I also think it's totally fine to fold it together and not have a custom error message. Check the implementations but I recall the validate one being written to throw if not valid.

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -74,6 +94,15 @@ const CreateWPForm: React.FC = () => {
     }
   };
 
+  const handleSubmit = (e: any) => {
+    e.preventDefault();
+
+    // exits handleSubmit if form input invalid (temporary)
+    if (!validate(e)) return;
+
+    console.log('form input validated');
```
jamescd18, line 103: Remove console.log before ready to merge

jamescd18 (changes requested):
This looks great to me!! Just a couple tweaks and cleanup. That's all the validation I think the form needs.

jamescd18 (approved):
LGTM!

Issue #468, opened 2022-01-21
Change Requests - Wire Up Create New Work Package Form


jamescd18:
## Description

The front-end visuals and backend endpoint for creating new work packages have been created in #466 and #364. This ticket is to complete the feature of creating new work packages by wiring up the front-end form to the endpoint and have it properly send data to the associated endpoint.

## Background

Wiring up the new work package form to the created endpoint is the last step in completing the feature for creating new work packages. See #448 for an example wiring ticket pull request.

## Prerequisites 

These tickets need to be completed before working on the current ticket:
- [x] #466 
- [x] #364 

## Proposed Solution

Gather together data from the front-end form into state management storage in the container component. Then facilitate packaging and sending the data to the back-end by:

- [x] writing the api function
- [x] writing the custom hook (using `useMutation`)
- [x] implementing the custom hook  (using the payload schema in the `utils` package)

Ensure all aspects of the work package create feature are present and functioning. Verify by tracing data from user input through the full stack to database ingestion. Implement / fix / write tests for any of the following elements in the trace through the tech stack that are not behaving as expected.

- [ ] Save/Cancel buttons
- [ ] Form input fields
- [ ] Create custom mutation hook for sending data to the backend 
- [ ] Create API function for sending data to backend
- [ ] Integrate custom mutation hook and API into front-end
- [ ] Api route string
- [ ] Api route function
- [ ] Payload validation
- [ ] Database insertion
- [ ] Api response

PR #527, opened 2022-02-15
#468 wire up create wp form

jpwaves:
Need review, will clean up chicken scratch code when I get back from work. Still need to fix `work-package-create` tests to conform to the new schema that uses WBS Numbers instead of project ID.

### DIFF
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -18,7 +18,7 @@ export const createWorkPackage: Handler = async ({ body }, _context) => {
   // and what number work package this should be
   const project = await prisma.project.findUnique({
     where: {
-      projectId: body.projectId
+      projectId: body.projectId // use single project as reference for breaking down wbsnum
     },
     include: {
       wbsElement: true,
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -10,7 +10,7 @@ import validator from '@middy/validator';
 import { Handler } from 'aws-lambda';
 import { PrismaClient, WBS_Element } from '@prisma/client';
 import { FromSchema } from 'json-schema-to-ts';
-import { buildSuccessResponse, eventSchema, WbsNumber, workPackageCreateInputSchemaBody } from 'utils';
+import { buildSuccessResponse, eventSchema, workPackageCreateInputSchemaBody } from 'utils';
 
 const prisma = new PrismaClient();
 
@@ -36,9 +36,9 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
   const wbsElem = await prisma.wBS_Element.findUnique({
     where: {
       wbsNumber: {
-        carNumber: carNumber,
-        projectNumber: projectNumber,
-        workPackageNumber: workPackageNumber
+        carNumber,
+        projectNumber,
+        workPackageNumber
       }
     }
   });
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -33,7 +33,7 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
   // and what number work package this should be
   const { carNumber, projectNumber, workPackageNumber } = projectWbsNum;
 
-  if (workPackageNumber === 0) throw new TypeError('Given WBS Number is not for a project.');
+  if (workPackageNumber !== 0) throw new TypeError('Given WBS Number is not for a project.');
 
   const wbsElem = await prisma.wBS_Element.findUnique({
     where: {
@@ -42,21 +42,20 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
         projectNumber,
         workPackageNumber
       }
+    },
+    include: {
+      project: {
+        include: {
+          wbsElement: true,
+          workPackages: { include: { wbsElement: true, dependencies: true } }
+        }
+      }
     }
   });
 
   if (wbsElem === null) throw new TypeError('No corresponding WBS Element for WBS Number.');
 
-  const project = await prisma.project.findUnique({
-    where: {
-      wbsElementId: wbsElem.wbsElementId
-      // use single project as reference for breaking down wbsnum
-    },
-    include: {
-      wbsElement: true,
-      workPackages: { include: { wbsElement: true, dependencies: true } }
-    }
-  });
+  const { project } = wbsElem;
 
   if (project === null) throw new TypeError('Project Id not found!');
   const { projectId } = project;
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -79,9 +79,10 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
     })
   );
 
-  if (dependenciesWBSElems === null) throw new TypeError('One of the dependencies was not found.');
-
-  const dependenciesIds = (dependenciesWBSElems as WBS_Element[]).map((ele) => ele.wbsElementId);
+  const dependenciesIds = dependenciesWBSElems.map((elem) => {
+    if (elem === null) throw new TypeError('One of the dependencies was not found.');
+    return elem.wbsElementId;
+  });
 
   console.log('creating');
   // add to the database
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -8,7 +8,7 @@ import jsonBodyParser from '@middy/http-json-body-parser';
 import httpErrorHandler from '@middy/http-error-handler';
 import validator from '@middy/validator';
 import { Handler } from 'aws-lambda';
-import { PrismaClient, WBS_Element } from '@prisma/client';
+import { PrismaClient } from '@prisma/client';
 import { FromSchema } from 'json-schema-to-ts';
 import { buildSuccessResponse, eventSchema, workPackageCreateInputSchemaBody } from 'utils';
 
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -32,7 +32,7 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
   // get the corresponding project so we can find the next wbs number
   // and what number work package this should be
   const { carNumber, projectNumber, workPackageNumber } = projectWbsNum;
-  console.log('wbs find unique');
+
   const wbsElem = await prisma.wBS_Element.findUnique({
     where: {
       wbsNumber: {
@@ -44,8 +44,7 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
   });
 
   if (wbsElem === null) throw new TypeError('No corresponding WBS Element for WBS Number.');
-  console.log('project find unique');
-  console.log(`wbsElemId: ${wbsElem.wbsElementId}`);
+
   const project = await prisma.project.findUnique({
     where: {
       wbsElementId: wbsElem.wbsElementId
@@ -84,7 +83,6 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
     return elem.wbsElementId;
   });
 
-  console.log('creating');
   // add to the database
   const created = await prisma.work_Package.create({
     data: {
@@ -113,7 +111,6 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
     }
   });
 
-  console.log('done');
   return buildSuccessResponse(created);
 };
 
--- a/src/backend/functions/work-packages-create.ts
+++ b/src/backend/functions/work-packages-create.ts
@@ -48,7 +48,7 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async 
   console.log(`wbsElemId: ${wbsElem.wbsElementId}`);
   const project = await prisma.project.findUnique({
     where: {
-      wbsElementId: wbsElem!.wbsElementId
+      wbsElementId: wbsElem.wbsElementId
       // use single project as reference for breaking down wbsnum
     },
     include: {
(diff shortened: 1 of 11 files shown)


jpwaves:
will fix merge conflicts when i get back from co-op as well

src/components/projects/create-wp-form/create-wp-form/create-wp-form.tsx
```diff
@@ -41,24 +44,57 @@ const CreateWPFormView: React.FC<CreateWPFormViewProps> = ({
...
                     <Form.Label htmlFor='duration'>Duration</Form.Label>
                     <InputGroup>
-                      <Form.Control id='duration' name='duration' aria-label={'duration'} type='number' min={0} required></Form.Control>
+                      <Form.Control
+                        id='duration'
+                        name='duration'
+                        aria-label={'duration'}
+                        type='number'
+                        min={0}
+                        onChange={(e: any) => duration.setter(e.target.value)}
```
acd124, line 96: For all of these I believe the `e: any` isn't necessary since react provides the types, but for the number inputs you might want to parse the values so it's not setting them as strings. I'm not sure about this though and I haven't tested it.

acd124 (changes requested):

src/components/projects/create-wp-form/create-wp-form/create-wp-form.tsx
jpwaves: Right now, my `handleSubmit` expects all the data to come in as strings, and it seems to work so far. I know we have plans to switch to using react forms in the future, so I'm just gonna leave it alone since it works for now and keep the number inputs as strings.

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -75,20 +98,88 @@ const CreateWPForm: React.FC = () => {
...
+      }
+      const depWbsNums = dependencies.map(dependency => {
+        const depWbsNum = validateWBS(dependency.trim());
+        return {
+          carNumber: depWbsNum.car,
+          projectNumber: depWbsNum.project,
+          workPackageNumber: depWbsNum.workPackage
+        }
+      });
+      if (depWbsNums) {
```
acd124, line 138: The typings say that `depWbsNums` will always result in an array, so it will always be truthy and the else won't be reached, not sure what you're testing here

src/utils/src/types/api-utils-types.ts
```diff
@@ -42,21 +42,21 @@ export const enumType = <Items extends Readonly<string[]>>(...items: Items) =>
...
+      type: 'object',
+      properties: {
+        'user-agent': stringType
+        // Cookie: stringType
+      },
+      required: ['user-agent']
+    }
+  },
+  required: ['body', 'headers']
+} as const);
```
acd124, line 59: I think the formatting here will go back if you save it again with prettier, mind if you check some of the files you made changes to?

acd124 (changes requested):

src/backend/functions/work-packages-create.ts
```diff
@@ -31,32 +31,67 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async
...
+            carNumber: ele.carNumber,
+            projectNumber: ele.projectNumber,
+            workPackageNumber: ele.workPackageNumber
+          }
+        }
+      });
+    })
+  );
+
+  if (dependenciesWBSElems === null) throw new TypeError('One of the dependencies was not found.');
```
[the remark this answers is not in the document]
jpwaves: Oh gotcha, that should be an easy fix

src/utils/src/types/api-utils-types.ts
jpwaves: Yeah, I can go check. I think there's also a command that does all the prettier fixes for you (`prettier ./src --fix` I think? Or something like that) which we should consider adding to the `package.json` scripts so that it's easier to fix these prettier issues in the future. I think there's also one for ESLint, but I'd have to check

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -75,20 +98,88 @@ const CreateWPForm: React.FC = () => {
...
     // exits handleSubmit if form input invalid (should be changed in wire up)
-    const wbsNum: WbsNumber = validateWBS(e.target.wbsNum.value.trim());
+    let wbsNum: WbsNumber;
+    try {
+      wbsNum = validateWBS(projectWbsNum);
 
-    if (!isProject(wbsNum!)) {
-      alert('Please enter a valid Project WBS Number.');
-      return;
+      const { userId } = authContext!.user!;
```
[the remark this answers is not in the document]
jpwaves: Found the hook, I'll add it in

src/utils/src/types/api-utils-types.ts
acd124: Yeah, great idea, I submitted a ticket for that

acd124 (commented):
I don't think I have any more issues with the code that I'd like to bring up. I know this does affect the front end, so I'm not sure if we can actually see the changes in action, or if that's a not yet for this ticket. Other than that I think it's just the `console.log`s that would get removed unless there are more changes to be made

src/backend/functions/work-packages-create.ts
```diff
@@ -31,32 +31,66 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async
...
+        carNumber,
+        projectNumber,
+        workPackageNumber
+      }
+    }
+  });
+
+  if (wbsElem === null) throw new TypeError('No corresponding WBS Element for WBS Number.');
+
   const project = await prisma.project.findUnique({
```
jamescd18, line 34: I think this query is unneeded because we can just add something like "include: project: true" to the lookup via WBS number. Then we'd just want to add a check before that to ensure that the supplied work package number is 0 to ensure that resulting project field on the WBS element will not be null

src/backend/tests/work-packages-create.test.ts
```diff
@@ -13,14 +13,20 @@ describe('work package create', () => {
     };
 
     describe('validate inputs', () => {
+      const mockWbsNum = {
```
jamescd18, line 16: I'm pretty sure there's a few of these in the test data you can use

jamescd18: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/src/test-support/test-data/wbs-numbers.stub.ts

src/backend/tests/work-packages-create.test.ts
```diff
@@ -13,14 +13,20 @@ describe('work package create', () => {
...
       const goodBody = {
         userId: 1,
         name: 'name',
         crId: 2,
-        projectId: 1,
+        projectWbsNum: mockWbsNum,
         startDate: '2015-10-06',
         duration: 1,
-        wbsElementIds: [1],
+        dependencies: [mockWbsNum],
```
jamescd18, line 29: New separate ticket cuz this reminded me of additional edge case business logic to disallow https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/535

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -3,23 +3,46 @@
...
 
 export interface EditableTextInputListUtils {
   add: (val: any) => void;
   remove: (idx: number) => void;
   update: (idx: number, val: any) => void;
 }
 
+export interface StateBundle {
+  state: any;
+  setter: (val: any) => void;
```
jamescd18, line 22: Pretty sure there's a better definition for this type

jamescd18: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/src/components/change-requests/review-change-request/review-change-request/review-change-request.tsx#L15

jamescd18: Also this interface belongs elsewhere probably. Maybe in the shared folder in a new file? Or something along the lines of component supports tools?

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -75,20 +98,83 @@ const CreateWPForm: React.FC = () => {
     }
   };
 
-  const handleSubmit = (e: any) => {
+  const stateBundleBuilder = (state: any, setter: any) => {
```
jamescd18, line 101: Better types, also move to diff file

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -75,20 +98,83 @@ const CreateWPForm: React.FC = () => {
...
+          projectNumber: wbsNum.project,
+          workPackageNumber: wbsNum.workPackage
+        },
+        startDate,
+        duration,
+        dependencies: depWbsNums,
+        expectedActivities,
+        deliverables
+      });
+      history.push(`${routes.CHANGE_REQUESTS}`);
```
jamescd18, line 153: If not modifying the route, drop the template string and just have the variable

src/components/projects/create-wp-form/create-wp-form/create-wp-form.test.tsx
```diff
@@ -3,24 +3,36 @@
...
-const mockDeliverables = ["go vroom"];
+const mockDependencies = ['1.2.3', '1.2.1', '1.12.0'];
+const mockEA = ['yooooo', 'wassup', 'bruh', 'whoosh'];
+const mockDeliverables = ['go vroom'];
 const mockUtils = {
   add: () => null,
   remove: () => null,
   update: () => null
 };
+const mockStateBundle = {
```
jamescd18, line 17: Pretty sure these can go in the proper test support utils stub file https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/src/test-support/test-data/test-utils.stub.ts

jamescd18 (changes requested):
This is looking really good! There's a few different changes I've requested, after which I think it'll ready for final review. Amazing job Justin!! This is definitely a tough one doing modifications to the back-end, but I think you've done a great job learning and adapting!

src/components/projects/create-wp-form/create-wp-form.tsx
jpwaves: Yeah, I realized after typing that out that I didn't actually need the state at all, so I'm going to remove that interface and just use the type from your referenced example.

src/components/projects/create-wp-form/create-wp-form.tsx
jpwaves: Function is no longer needed after removing the `StateBundle` interface

src/backend/functions/work-packages-create.ts
```diff
@@ -31,32 +31,67 @@ export const createWorkPackage: Handler<FromSchema<typeof inputSchema>> = async
...
+        projectNumber,
+        workPackageNumber
+      }
+    },
     include: {
-      wbsElement: true,
-      workPackages: { include: { wbsElement: true, dependencies: true } }
+      project: {
+        include: {
+          wbsElement: true,
```
jamescd18, line 49: WbsElement is the parent of project, so by doing `include` within `project`, you’re saying get A > get B > get A and duplicating pulling the wbsElement

src/components/projects/create-wp-form/create-wp-form/create-wp-form.test.tsx
```diff
@@ -3,16 +3,19 @@
...
-const mockDeliverables = ["go vroom"];
-const mockUtils = {
-  add: () => null,
-  remove: () => null,
-  update: () => null
+import { mockUtils } from '../../../../test-support/test-data/test-utils.stub';
+import { render, screen } from '../../../../test-support/test-utils';
+import CreateWPFormView from './create-wp-form';
+
+const mockDependencies = ['1.2.3', '1.2.1', '1.12.0'];
```
jamescd18, line 10: This could probably just pull from the wbsNumber stub file

jamescd18 (changes requested):

src/backend/functions/work-packages-create.ts
jpwaves: Ah, I see. I'll remove that from the include

src/components/projects/create-wp-form/create-wp-form.tsx
```diff
@@ -75,20 +93,76 @@ const CreateWPForm: React.FC = () => {
...
+        startDate,
+        duration,
+        dependencies: depWbsNums,
+        expectedActivities,
+        deliverables
+      });
+      history.push(routes.CHANGE_REQUESTS);
+    } catch (e) {
+      console.log(e);
+      alert('wbs num not valid');
```
jamescd18, line 144:
```suggestion
      alert('something went wrong');
```

jamescd18: this is because this catches any errors from the back-end call and so it won't always be because of the wbs nums

src/components/projects/create-wp-form/create-wp-form/create-wp-form.tsx
```diff
@@ -40,34 +43,72 @@ const CreateWPFormView: React.FC<CreateWPFormViewProps> = ({
...
+                      onChange={(e) => crId(parseInt(e.target.value))}
+                      required
+                    ></Form.Control>
                   </Form.Group>
                 </Row>
                 <Row>
                   <Form.Group as={Col}>
-                    <Form.Label htmlFor='start-date'>Start Date</Form.Label>
-                    <Form.Control id='start-date' name='startDate' aria-label={'start date input'} type='date' required></Form.Control>
+                    <Form.Label htmlFor="start-date">Start Date</Form.Label>
```
jamescd18, line 81:
```suggestion
                    <Form.Label htmlFor="start-date">Start Date (YYYY-MM-DD)</Form.Label>
```

jamescd18: this is because it will error with different formats that people might think of using

jamescd18 (commented):
this looks soooooo gooooood!!! and it works!!!! aaaahhhhhhhh!!! yaaaayy!!!

jamescd18 (approved):

acd124 (approved):
LGTM 👍🏼

Issue #477, opened 2022-01-21
General - Standardize Date Time Display


jamescd18:
## Description

The variety of different date formats across the application makes it confusing for users, forcing them to read less-readable dates and mentally adjust to read the different date formats.

## Current Behavior

There are a variety of different date display formats across the application. See screenshots for details.

## Desired Behavior

Standardize all date displaying parts of the application front-end to utilize the `datePipe`. Ensure the date pipe is displaying dates as `MM/DD/YYYY` where all single digits are padded with a zero. Examples: `01/04/2022`, `11/29/2021`.

## Screenshots

<img width="194" alt="Screen Shot 2022-02-02 at 11 05 51 AM" src="https://user-images.githubusercontent.com/51571627/152191141-30cc606f-75a2-49fb-86da-d41f354a26e4.png">
<img width="175" alt="Screen Shot 2022-02-02 at 11 05 41 AM" src="https://user-images.githubusercontent.com/51571627/152191114-2563cf4d-5577-4e42-8e8e-79d19eda2d9f.png">
<img width="201" alt="Screen Shot 2022-02-02 at 11 06 11 AM" src="https://user-images.githubusercontent.com/51571627/152191194-87ffa5a1-3a48-4fdc-ab13-2e605c221e0a.png">
<img width="209" alt="Screen Shot 2022-02-02 at 11 06 18 AM" src="https://user-images.githubusercontent.com/51571627/152191216-8ae1b2e6-d644-44b0-b92a-787c91e1389d.png">
<img width="170" alt="Screen Shot 2022-02-02 at 11 06 35 AM" src="https://user-images.githubusercontent.com/51571627/152191269-e36775c5-c8f5-495b-bdbe-7514bb95690e.png">
<img width="284" alt="Screen Shot 2022-02-02 at 11 06 45 AM" src="https://user-images.githubusercontent.com/51571627/152191298-c8e68b50-6c2f-40dd-aece-a20b4cb4c376.png">
<img width="292" alt="Screen Shot 2022-02-02 at 11 07 01 AM" src="https://user-images.githubusercontent.com/51571627/152191348-fcc5de8e-522a-4f82-b08e-31d065e07c29.png">
<img width="251" alt="Screen Shot 2022-02-02 at 11 07 24 AM" src="https://user-images.githubusercontent.com/51571627/152191430-f06393b2-90d8-41aa-a5fb-866168006c45.png">

PR #529, opened 2022-02-17
#477 Standardized Date Format

oneal55:
Applied datePipe wherever it was needed


acd124:
I know you haven't requested a review yet, but I did want to comment on a few things from the [Contributor guide](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/docs/ContributorGuide.md) that you might want to look at going forward. You'll want to be including the issue that each commit is for in the commit message, in this case that's #477, so an example commit message for this might be `#477 - always use datePipe` or something similar, there's tips about the format in the contributor guide. Another thing is I see is that you have the `package-lock.json` and a file from `.vscode/` included in the commit. Generally we only want to be committing the `package-lock.json` when we make changes to the dependencies, so in this case where there are no dependency updates, we would restore the previous version of the `package-lock.json` and leave it out of the commit. As for the `launch.json` I'm not sure if we want to add it the `.gitignore` or not, but I'm fairly sure this ticket doesn't require it being added, so I don't think we want to commit it to the repository just yet. Feel free to reach out in the slack software channel to talk more about this in a more conversation style, and other than these notes I do think it's done pretty well.

Issue #485, opened 2022-01-21
Change Requests - "Type" input has incorrect option 


jpwaves:
### Description

There needs to be some fixes for the options avalable for the type of change request.

### Browser(s)

Firefox

### Platform(s)

Desktop

### Version(s)

main

### Observed Behavior

In the dropdown for selecting the type of a change request "Initiation" is an option. However, there is no "Initiation" type change request. "Activation" change requests, however, are missing from the dropdown.

### Proposed Solution
"Activation" should be an option in the dropdown and "Initiation" should not.

Ensure that the values in the dropdown are generated from `Object.values(ENUM)` where `ENUM` is the proper enum for the values in the dropdown.

### Steps to Reproduce

1. Go to new change request form
2. Select dropdown for the "Type" input

![new-cr-initiation-type](https://user-images.githubusercontent.com/48561685/150611889-1f06fcb6-0a9f-4cbc-8eca-154f3ac48854.png)

usayd-sheikh:
#570 Merged PR

PR #534, opened 2022-02-19
#485 update new cr fields

syedDS:
new C_R update from initiation to activation on type

### DIFF
--- a/src/components/change-requests/new-change-request/new-change-request-page/common-form-fields/common-form-fields.tsx
+++ b/src/components/change-requests/new-change-request/new-change-request-page/common-form-fields/common-form-fields.tsx
@@ -8,6 +8,7 @@ import { exampleAllWorkPackages } from '../../../../../test-support/test-data/wo
 import { exampleAllProjects } from '../../../../../test-support/test-data/projects.stub';
 import { wbsPipe } from '../../../../../shared/pipes';
 import './common-form-fields.module.css';
+import {ChangeRequestType}  from 'utils';
 
 const CommonFormFields: React.FC = () => {
   return (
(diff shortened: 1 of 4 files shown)


syedDS:
The code tested on the local repo didnt gave any error but is giving after PR , can someone review it

jamescd18:
It looks like you changed the `package.json` and `package-lock.json` files, both of which you do not want to be changing. This potentially is part of why it broke.

src/components/change-requests/new-change-request/new-change-request-page/common-form-fields/common-form-fields.tsx
```diff
@@ -35,7 +36,9 @@ const CommonFormFields: React.FC = () => {
       <div className={'px-4'}>
         Type
         <Form.Control as="select" custom>
-          {['New Function', 'Design Issue', 'Other', 'Initiation', 'Stage Gate'].map((t) => (
+          {Object.values(ChangeRequestType).map((
+            t // change from intiation -> Activation
```
jamescd18, line 40: This comment looks like it's specific to this pull request so you'll want to remove it

jamescd18 (changes requested):
Awesome! Just one small change for this code. The bigger thing is that this PR has a very messy commit history and I don't see any of the commits being tagged with the ticket number. Given how small the code change is, I would love for you to create a fresh new branch, make just 1 commit that has the ticket number as described in the contributor guide, and open a new PR. I think this is a really good learning experience because this PR here allowed you to figure out how things work and get feedback, and then the fresh new PR will be the one that we actually approve and merge into main. If you have any questions or if anything is confusing, reach out to your tech lead on Slack. Thank you!

Issue #540, opened 2022-02-21
Projects - Create Work Package Can Be Performed With Unreviewed / Denied Change Request


jamescd18:
## Description

Make sure only change requests IDs that correspond to reviewed and approved change request IDs can be submitted on the create new work package form.

## Browser(s)

Chrome, Firefox, Safari

## Platform(s)

Desktop, iOS, Android

## Version(s)

main

## Observed Behavior

You can submit the create new work package form (or to the endpoint) and supply a change request ID that corresponds to an unreviewed change request or a denied change request and it will work

## Expected Behavior

The back-end endpoint for submitting new work packages only accepts change request IDs for change requests that are reviewed and approved. This matches the back-end endpoint for creating new projects. 

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/d7ab161dc51d936a5cb5a9ac3f05c40cfaee3b25/src/backend/functions/projects-new.ts#L34

### Steps to Reproduce

1. Have an unreviewed change request, denied change request, and accepted change request
2. Go to the create new work package form from the accepted change request
3. Put in the change request ID of the unreviewed change request or the denied change request
4. Fill in the rest of the form with normal data
5. Submit and notice that it goes through
6. Observe the unreviewed / denied change request who's ID was submitted and see how it is not properly accepted but still gets a change logged and status changed to implemented

### Screenshots (as needed)

<img width="1214" alt="Screen Shot 2022-02-21 at 2 21 00 AM" src="https://user-images.githubusercontent.com/51571627/154907277-ce7175ec-391d-40b5-8289-9b91c849eb36.png">

<img width="1386" alt="Screen Shot 2022-02-21 at 2 21 14 AM" src="https://user-images.githubusercontent.com/51571627/154907309-cdf57fd7-66d3-4020-99b8-59a72b64f3c4.png">

PR #542, opened 2022-02-21
#467 wire up create project form

jpwaves:
Fairly straight forward, made the api call, custom react hook, and used states to connect the form data to `handleSubmit`. For future wire up tickets, I'd suggest adding in the tickets to try to use the available payload schemas we now have as that significantly simplifies the code.


jamescd18 (approved):
LGTM. Created #543 and #540 for these forms tho.

PR #548, opened 2022-02-21
Add loading indicator to create project form and create work package form

### LINKED ISSUE #543: Projects - Add Loading Indicator to Create Forms
### Desired Changes

Implement the pattern of using the loading indicator in the create new project form and create new work package form to better communicate to users when the application is loading and working on things instead of doing nothing.

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/d7ab161dc51d936a5cb5a9ac3f05c40cfaee3b25/src/components/projects/projects-table/projects-table.tsx#L71

### Screenshots (as needed)

_No response_

jamescd18:
Pretty self-explanatory. Minor user experience improvement. Manually tested and it works. Closes #543.

### DIFF
--- a/src/components/projects/create-project-form/create-project-form.tsx
+++ b/src/components/projects/create-project-form/create-project-form.tsx
@@ -8,6 +8,7 @@ import { useHistory } from 'react-router-dom';
 import { useAuth } from '../../../services/auth.hooks';
 import { useCreateSingleProject } from '../../../services/projects.hooks';
 import { routes } from '../../../shared/routes';
+import LoadingIndicator from '../../shared/loading-indicator/loading-indicator';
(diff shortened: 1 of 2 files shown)


jpwaves (approved):
lgtm, also tested manually on my local and everything worked as expected. not sure if it's necessary, but we could look into doing more thorough testing with transition states like these in the future so we can have jest tests to check our loading indicator appears, in addition to manual testing.

jamescd18:
Yeah I think I initially didn't bother testing loading states because there's no text in the screen to grab and because I think we may eventually want a spike for non-full-page loading spinners (specific blocks, or overlays?) so yeah testing via a test ID would probably be good

Issue #417, opened 2022-01-03
General - Create New NPM Script


jinhyunh:
## Description

Create a new npm script as an alternative to `npm install` that does not build the dependencies included with `npm install`. 

## Background

Currently, we have the `npm install` script which runs `npm prisma generate`, the utils package re-generation, and fetches remote dependent packages.

When you call `npm install` currently, the `preinstall` is called first which calls `npm run repack-utils-pkg` which calls `cd src/utils && npm i && npm run build`. This function calls `rimraf lib` and `tsc` in utils as a part of their `prebuild` and `build`. As a part of the `postinstall` of `npm install`, `npm run prisma:generate` is called and as a part of that command, `npx prisma generate` is called.

## Desired Additional Features

We would like to make a new script in the package.json that runs just the `npm prisma generate` and the utils package re-generation. This would let developers run this script instead rather than running `npm install`. Documentation explaining this script and what it does is needed as well.

## Proposed Solution

- [ ] Create a new script in `pacakge.json` that solely runs `npm prisma generate` and utils package re-generation
- [ ] Documentation for this script explaining what the script does and why/when to run it

PR #550, opened 2022-02-23
#417 Create npm script

VatsalMehta27:
Created a script named "light-install" that runs the "preinstall" and "postinstall" scripts. 

We could use some clarification on the ticket about why/when a dev would want to run the script, so the documentation isn't done yet.

Ticket: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/issues/417


jamescd18 (changes requested):
This is a great start! Definitely the right direction to create a new "light-install" script. There's a few changes here that I think would be best resolved by creating a fresh new pull request, given how the change is just 1 line.

- we should not be committing changes to the package-lock.json because we're not making any changes to the dependencies within the package.json
- all commits should be tagged with the issue number
- the light-install script also needs to install the local utils package in order to propagate changes from within the local package to the `node_modules` folder

Please let me know if there's any questions or issues, and ping me again on Slack when you've got things ready for review again. Thank you!

Issue #411, opened 2021-12-29
Projects - Properly Utilize Backend Support Functions


jamescd18:
## Description

The `src/utils/src/backend-supports` folder was created to abstract common functions performed by the backend into the utils package as per what is beneficial, functional, and feasible. The backend supports were added, but never utilized as much as they should be, so let's change that.

## Proposed Solution

- [x] Rename the `src/utils/src/backend-supports/projects-get-all.ts` file to something more generalized like `project-supports.ts`
- [x] Rename the corresponding test file too away from `src/utils/tests/all-projects-support.test.ts`
- [x] Redirect and abstract all backend lambda endpoints to utilize the end date calculation function
- [x] Redirect and abstract all backend lambda endpoints to utilize the duration calculation function

Feel free to include edits to the existing abstracted function implementations in the pull request you make, and propose any additional functions that should be abstracted. Keep in mind that the abstracted functions should be thoroughly tested and must be imported in the application via the `utils` package. 

Also remember that you need to run `npm install` every time you've made changes to the utils code in order for it to be re-bundled and made available via importing `utils`.

PR #551, opened 2022-02-23
#411 utilize backend funcs

lunazzao:
#411 - Renamed the file with "project-supports.ts", 
Replaced two lambda functions from "src/backend/functions/work-packages.ts(line 73)" and "src/backend/functions/projects.ts(line 130)"  with the one in utils, added comments for both functions, and changed the second name to "calculateDuration", Fix usage in "indecx.ts"

So far don't have recommendations for other lambdas.


src/backend/functions/work-packages.ts
```diff
@@ -69,8 +70,10 @@ const workPackageTransformer = (
   if (payload === null) throw new TypeError('WBS_Element not found');
   const wbsElement = 'wbsElement' in payload ? payload.wbsElement : payload;
   const workPackage = 'workPackage' in payload ? payload.workPackage! : payload;
-  const endDate = new Date(workPackage.startDate);
-  endDate.setDate(workPackage.duration * 7);
+  //const endDate = new Date(workPackage.startDate);
```
anthonybernardi, line 73: you can delete these comments

src/utils/src/backend-supports/project-supports.ts
```diff
@@ -3,13 +3,24 @@
...
 const calculateEndDate = (start: Date, weeks: number) => {
   const end = new Date(start);
   end.setDate(start.getDate() + weeks * 7);
   return end;
 };
 
-const projectDurationBuilder = (wps: any) => {
+/**
+ * This symbol function calculates the duration 
+ * @param wps an array of a project schedule
```
anthonybernardi, line 20: this is an array of work packages

src/backend/functions/work-packages.ts
```diff
@@ -21,6 +21,7 @@ import {
   WorkPackage,
   WbsElementStatus
 } from 'utils';
+import { calculateEndDate, projectDurationBuilder } from 'utils';
```
anthonybernardi, line 24: take out projectDurationBuilder to get rid of this lint error

src/utils/src/backend-supports/project-supports.ts
```diff
@@ -26,4 +37,4 @@ const projectDurationBuilder = (wps: any) => {
   return Math.round(durationWeeks);
 };
 
-export { projectDurationBuilder, calculateEndDate };
+export { calculateDuration as projectDurationBuilder, calculateEndDate };
```
anthonybernardi, line 40: the refactoring just exported this as what it was before so it didn't really do anything. look into refactoring with VSCode so it refactors the whole thing

anthonybernardi:
One last thing: the tests are failing because the import statement in projects-supports.test.ts has not been changed to match the new file name. Also the test file is 'projectS-supports' and the actual file is 'project-supports' so that should be fixed to be consistent too

src/utils/src/backend-supports/project-supports.ts
lunazzao: image
if we change it array of Array<WorkPackage>, it would fail the pre-written tests in src/utils/tests/projects-supports.test.ts, change anyways?

anthonybernardi: I just meant to change the comment to say 'wps is an array of work packages', and yeah good point just leave it as 'any' type in the arguments for now.

anthonybernardi:
Great work! Almost finished, just a few more things:
- the main file is named 'project-supports.ts' and the test file is named 'projects-supports.test.ts'. The plurals don't match up (project vs projects) so I would change the test file name to be 'project-supports.test.ts' instead so that it's consistent
- the commits for this PR don't follow the required naming convention (e.g., "#411 - added something"). You'll have to redo the PR with these guidelines followed. I laid out the steps for this in the slack but let me know if you have questions about that

PR #552, opened 2022-02-25
#541 - Supply correct wbsNum to implemented changes

### LINKED ISSUE #541: Change Requests - Implemented Changes List Has Incorrect WBS Numbers
### Browser(s)

Firefox

### Platform(s)

Desktop

### Version(s)

main

### Observed Behavior

The linked WBS numbers for each implemented change displays the incorrect WBS number for some projects and work packages.

### Expected Behavior

The WBS numbers linked for each implemented change should route to the correct project/work package and also display the correct WBS number for each as well.

### Steps to Reproduce

1. Create a new work package or project
2. Go to the change request details page associated with the newly created work package/project
3. Compare the WBS Numbers in the Implemented Changes list to those stored in your database

### Technical Notes

Code in question is located in `backend/functions/change-requests.ts`, specifically these lines:
https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/d7ab161dc51d936a5cb5a9ac3f05c40cfaee3b25/src/backend/functions/change-requests.ts#L73-L76
We should not be reusing the `wbsNum` we defined previously here as that refers to the WBS Element that the change request is linked to, which is different from the `wbsNum` for each individual change (each change could refer to a different work package or project than the change request itself is linked to).

### Screenshots (as needed)

bugged-wbsnums1
bugged-wbsnums2

jamescd18:
Fixes #541 

we were just supplying the wrong WBS num in the back-end. now it works beautifully

Screen Shot 2022-02-25 at 11 39 44 AM

### DIFF
--- a/src/backend/functions/change-requests.ts
+++ b/src/backend/functions/change-requests.ts
@@ -30,7 +30,8 @@ const relationArgs = Prisma.validator<Prisma.Change_RequestArgs>()({
     reviewer: true,
     changes: {
       include: {
-        implementer: true
+        implementer: true,
+        wbsElement: true
       }
     },
     scopeChangeRequest: { include: { why: true } },
@@ -72,7 +73,11 @@ const changeRequestTransformer = (
     ),
     implementedChanges: changeRequest.changes.map((change) => ({
       ...change,
-      wbsNum
+      wbsNum: {
+        car: change.wbsElement.carNumber,
+        project: change.wbsElement.projectNumber,
+        workPackage: change.wbsElement.workPackageNumber
+      }
     })),
     wbsNum,
     ...changeRequest.scopeChangeRequest,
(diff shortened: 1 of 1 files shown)


src/backend/functions/change-requests.ts
```diff
@@ -72,7 +73,11 @@ const changeRequestTransformer = (
     ),
     implementedChanges: changeRequest.changes.map((change) => ({
       ...change,
-      wbsNum
+      wbsNum: {
```
acd124, line 76: Since we're making multiple wbsNums here as well, I would copy the `wbsNumOf` function from another file like projects.ts, don't think it can go to utils yet since it has its arg type of WBS_Element which is from prisma, if that were changed it might go in utils.

jpwaves (dismissed):
i would look into alexander's comment to reduce repetitive code, but otherwise lgtm and it works fine when i tested on local.

Issue #406, opened 2021-12-28
General - Logout Button Not Working


RChandler234:
### Browser(s)

Chrome

### Platform(s)

Desktop

### Version(s)

other (specify below)

### Observed Behavior

When the logout white space around the logout button is clicked, nothing happens. The space around the "Logout" label does not trigger the button and log the user out.

### Expected Behavior

The logout button should log you out of the application and take you back to the login screen when clicked.

### Steps to Reproduce

1. Login to the application
2. Navigate to the Logout Button
3. Try to click the whitespace around the logout button

### Proposed Solution

Adjust the white space so that the click target covers more of the whole area, so that users are able to logout by clicking anywhere on the logout button. 

### Screenshots (as needed)

Menu item target area. Lots of this area is clickable but won't trigger the log out function.
<img width="227" alt="Screen Shot 2022-02-14 at 1 22 55 AM" src="https://user-images.githubusercontent.com/51571627/153811520-5060842a-70e6-4cbb-b13f-2dde38926c5b.png">

When the mouse hovers on the text itself the color changes and this text is the area that when clicked actually triggers the log out function.
<img width="234" alt="Screen Shot 2022-02-14 at 1 23 10 AM" src="https://user-images.githubusercontent.com/51571627/153811658-7ab514a6-caa8-4123-8c46-1b37d96f9602.png">

Video screen recording
https://user-images.githubusercontent.com/51571627/153811441-16ab1343-2b2a-4d80-b3e6-244f824f6628.mov

jamescd18:
The logout button works perfectly for me on macOS Big Sur version 11.0.1 with Safari version 14.0.1

Let's get some more data points and see whether or not this works for others.

jamescd18:
It looks like this could be caused by the text of the button triggering the button, but the space around the text not triggering the button. If that is the case, then we likely need to adjust it so that the click target covers more of the whole area

PR #556, opened 2022-02-26
#406 Logout button now mirrors the setting button

MagicalLime:
Fixed the front-end functionality for the logout button. Issue was related to GoogleLogout causing weird spacing issues. Fixed by including a link item within the GoogleLogout render prop. 

Also, whilst fixing this, we found an issue that the chrome back and forward buttons would still allow you to evade the log out. Example: You login, you logout, you hit the chrome back button and you are on the same page as before logging out with full functionality as a user.

Seemed out of scope to fix that issue, but just wanted to point it out.

### DIFF
--- a/src/components/whole-app/nav-top-bar/nav-user-menu/nav-user-menu.tsx
+++ b/src/components/whole-app/nav-top-bar/nav-user-menu/nav-user-menu.tsx
@@ -31,23 +31,27 @@ const NavUserMenu: React.FC = () => {
         </Link>
       </NavDropdown.Item>
       <NavDropdown.Item as="div">
-        <GoogleLogout
-          clientId={process.env.REACT_APP_GOOGLE_AUTH_CLIENT_ID!}
-          //jsSrc={'accounts.google.com/gsi/client'}
-          onLogoutSuccess={() => {
-            auth!.signout();
-            history.push(routes.HOME);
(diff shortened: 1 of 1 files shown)


src/components/whole-app/nav-top-bar/nav-user-menu/nav-user-menu.tsx
```diff
@@ -31,23 +31,27 @@ const NavUserMenu: React.FC = () => {
...
-            <button
-              className={'nav-link p-0 m-0 ' + styles.logoutButton}
-              onClick={renderProps.onClick}
-              disabled={renderProps.disabled}
-            >
-              Logout
-            </button>
-          )}
-        ></GoogleLogout>
+        <Link className="nav-link p-0" role="button" to="/login">
```
jamescd18, line 34: No hard-coded strings. You'll need to reference the correct file with strings

jamescd18: As seen below on line 40

jamescd18 (changes requested):
This is a great start! Did you reproduce the evading log out bug on production? If not, it might be due to your fix.

jamescd18:
Closes #406

PR #558, opened 2022-03-02
#417 - Create New "light-install" NPM Script

LiamSmeraglia:
“light-install” is a lighter alternative to npm install that avoids building unnecessary dependencies. This can be used for local development to cut down on development time. The script works by first running preinstall, which builds the packages in /utils. It then installs them into the global node_modules, ending by running postinstall to generate prisma.


LiamSmeraglia:
Sorry I messed up the commit messages, didn't realize it needed the # to link. Will do in the future.

jamescd18 (approved):
Aight let's get this merged

PR #559, opened 2022-03-02
initial step for resetting

lunazzao:
#411 - utilize abstract common functions and rename files.


lunazzao:
Closed due to proper naming

PR #561, opened 2022-03-02
#367 project edit endpoint

### LINKED ISSUE #367: Projects - Project Edit Endpoint
## Description
See #368. It's going to have a ton of the same stuff, just with different fields.

lpfahler16:
Done(ish)! We are 98% of the way there. Following along with #368, I think everything is done as needed. I am getting some weird stuff from the description bullets though. First, description bullets require an id. I'm assuming we don't give ids to new description bullets (the database does) so how does that work? Also the helper method was listing all already deleted bullets as deleted again in the changes. I fixed that but am now getting some weird remove / add stuff in the change log. The description bullets overall seem kind of fishy to me so information on that would be great. Thanks!

closes #367


RChandler234:
hey logan, could you specifically reference what part of the code you're talking about in your comment, there should be a feature that lets you do this in github if you go to files changed; also are the issues you're talking about related to the jest tests that are failing? If not I'd take care of those first before doing anything else.

src/utils/src/types/project-types.ts
```diff
@@ -77,3 +77,80 @@ export const createProjectPayloadSchema = {
...
+    },
+    goals: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          id: { type: 'integer', minimum: 0 },
+          detail: { type: 'string' }
+        },
+        required: ['id', 'detail']
```
lpfahler16, line 105: This is a description bullet. Here, I have id required as it is in ticket #368

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,492 @@
...
+  // return the updated work package
+  return buildSuccessResponse(updatedProject);
+};
+
+// /**
+//  * HELPER METHODS:
+//  */
+
+// helper method to add the given description bullets into the database, linked to the given work package
+const addDescriptionBullets = async (
```
lpfahler16, line 295: Here, when the description bullets are added to the database, we do not use any id passed (only the id for the relation to the project)

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,492 @@
...
+      wbsElementId,
+      detail: `${element.type} ${nameOfField} "${element.element}"`
+    };
+  });
+};
+
+// this method creates changes for description bullet inputs
+// it returns it as an object of {deletedIds[], addedDetails[] changes[]}
+// because the deletedIds are needed for the database and the addedDetails are needed to make new ones
+export const createDescriptionBulletChangesJson = (
```
lpfahler16, line 398: This is the method that is giving me trouble. It originally would list any description bullet that had been deleted as deleted again. I fixed that by filtering out the deleted ones. Now it lists any that don't change as deleted and then added again.

lpfahler16:
Update: Here is my plan.

I would like to make the id field for description bullets optional.

If no id specified -> "added"

If id specified and text is different than before -> "edited"

If id of previous bullet does not appear in new list and bullet has not already been deleted -> "Deleted"

I am going to implement these changes with the next commit. This does differ from #368 so if this is a problem I can implement it a different way.

jamescd18:
That sounds like the correct behavior to me. How does it differ from #368?

lpfahler16:
> That sounds like the correct behavior to me. How does it differ from #368?

ID is required in #368 and checked against IDs in the database, even in the case of new items.

src/utils/src/types/project-types.ts
```diff
@@ -77,3 +77,79 @@ export const createProjectPayloadSchema = {
 } as const;
 
 export type CreateProjectPayload = FromSchema<typeof createProjectPayloadSchema>;
+
+export const projectEditInputSchemaBody = {
```
jamescd18, line 81: Is there any reason to not utilize the abstractions provided here? https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/main/src/utils/src/types/api-utils-types.ts

You might need to fast-forward your branch and fix merge conflicts to get the latest stuff from main. The abstractions should preserve all functionality and greatly simplify the code. It's being used in other payload schema definitions.

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,528 @@
...
+      wbsElement: true,
+      goals: true,
+      features: true,
+      otherConstraints: true
+    }
+  });
+
+  // if it doesn't exist we error
+  if (originalProject === null) {
+    throw new TypeError('Project with given wbsElementId does not exist!');
```
jamescd18, line 52: This should be using `return buildClientFailureResponse('msg');` instead of throwing a TypeError. I know it's just copy-pasting from existing code, but basically a) this isn't the correct scenario for a type error and b) throwing an error in the back-end makes for terrible and difficult to decipher error messages.

jamescd18 (changes requested):
Definitely making its way towards complete! I've left some thoughts and would love others' input

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,528 @@
...
+  //   }
+  // });
+
+  // newArray.forEach((element) => {
+  //   if (element.id < 0 || !seenOld.has(element.id)) {
+  //     changes.push({ element, type: 'Added new' });
+  //   } else if (seenOld.get(element.id) !== element.detail) {
+  //     changes.push({ element, type: 'Edited' });
+  //   }
+  // });
```
RChandler234, line 444: is there a reason that all this commented code is still in here? if not, I'd say go ahead and take it out

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,528 @@
...
+
+  // update the project with the input fields
+  const updatedProject = await prisma.project.update({
+    where: {
+      wbsElementId
+    },
+    data: {
+      budget,
+      summary,
+      googleDriveFolderLink: body.googleDriveFolderLink,
```
anthonybernardi, line 231: may as well just extract all of these at the beginning too. if they're undefined that's fine because it'll get set to undefined here

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,528 @@
...
+
+  // All elements that were inputs but are not new
+  const existingElements = new Map<number, string>();
+
+  // Database version of edited elements
+  const originalElements = new Map<number, string>();
+
+  // Find new elements
+  newArray.forEach((element) => {
+    if (element.id === undefined) {
```
anthonybernardi, line 462: This is different than how I did it for work packages, although this might be a better way tbh

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,528 @@
...
+  //     changes.push({ element, type: 'Edited' });
+  //   }
+  // });
+
+  // Changes
+  const changes: { element: DescriptionBullet; type: string }[] = [];
+
+  // Elements from database that have not been deleted
+  const oldArrayNotDeleted = oldArray.filter((element) => {
+    return element.dateDeleted === undefined;
```
anthonybernardi, line 451: This was not something I considered in the work packages edit which may turn out to be a bug there

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,528 @@
...
+  //   } else if (seenOld.get(element.id) !== element.detail) {
+  //     changes.push({ element, type: 'Edited' });
+  //   }
+  // });
+
+  // Changes
+  const changes: { element: DescriptionBullet; type: string }[] = [];
+
+  // Elements from database that have not been deleted
+  const oldArrayNotDeleted = oldArray.filter((element) => {
```
anthonybernardi, line 450:
```suggestion
  const oldArrayNotDeleted = oldArray.filter((element) => element.dateDeleted === undefined);
```

src/backend/functions/projects-edit.ts
lpfahler16: That creates some weird bugs. I tried it and the code just became more complicated.

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,501 @@
...
+          name,
+          status: wbsElementStatus,
+          projectLeadId: body.projectLead,
+          projectManagerId: body.projectManager
+        }
+      }
+    }
+  });
+
+  // --> Pointer
```
jamescd18, line 252: ?

src/backend/functions/projects-edit.ts
```diff
@@ -0,0 +1,501 @@
...
+  };
+};
+
+// expected structure of json body
+const inputSchema = {
+  type: 'object',
+  properties: {
+    body: projectEditInputSchemaBody
+  }
+};
```
jamescd18, line 494: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/fcb5215b796b9800bdf33a171ab01a52bd8ac396/src/backend/functions/change-requests-review.ts#L50

jamescd18 (changes requested):
This is amazing! Really great job on this one given how tough it is. There's a couple smaller items to change, but it looks good from my testing.

lpfahler16:
All set!

jamescd18 (approved):
LGTM!

Issue #530, opened 2022-02-18
General - Create Prettier Script


acd124:
### Description

Create a script to run prettier on all files, similar to the `lint` script used to run eslint on all files at once.

### Background

We use prettier to automatically format files so the code is easier to read and consistent. We recommend that everyone uses format on save to make sure that every file they work on is properly formatted, but that isn't and hasn't always been done, so it is useful to be able to check and/or fix all the formatting issues.

### Proposed Solution

An npm script very similar to the `lint` script the check all relevant files. Make sure that things like the docs (text files), and ignored files aren't accidentally included in the check. At the very least all `.ts` and `.tsx` files in the `src` folder should be checked, though there may be other files that warrant being checked.

jpwaves:
I think having two scripts, one to merely check if there are files that violate prettier format, and one that actually goes in to automatically fix them (`--fix` flag) would be useful.

PR #568, opened 2022-03-07
#530 add scripts to check and enforce prettier

jpwaves:
## Changes

Added 2 scripts to `package.json`, one for checking prettier on files and one for enforcing prettier formatting.

## Notes

I don't know which files/folders to have prettier to ignore. I'll need to add that into a prettier config file before finishing this pull request.

## Test Cases

None

None

## To Do

- [x] Create prettier config file for what files to ignore
- [x] Move prettier config options from `package.json` into a separate prettier config file

## Checklist

- [x] No linting errors
- [x] No merge conflicts
- [x] All checks passing

Closes # (issue #)


jamescd18 (commented):
Lol that's way simpler than expected. Looks like a good start tho. 

Because we don't have any prettier ignore patterns, I'm pretty sure prettier is applied to all files in the repo. The main idea here tho is that it is applied to ts, tsx, js, json, css, md, and html. Are there any issues with applying it to all files? Does it cause sweeping changes if you do? (Which would indicate we're not currently running it on all files)

Also, I don't care / don't know whether it's in this ticket or another ticket, this PR or another PR, but I want to see if we can get this script to be run as a GitHub Action check for all PRs, just the way we do with linting. This would dramatically improve our ability to easily enforce prettier formatting.

jamescd18 (changes requested):
Oops, yes, `src/utils/lib` is definitely generated and should be ignored for prettier. Definitely any generated files should be ignored. What other files are showing need changes? Requesting changes for these include/exclude files stuff. 

I'm not too keen on a pre-commit hook unless you think it's the proper way to do this. I have second-hand bad experience with pre-commit hooks from first co-op which is why I was suggesting trying to get it to run as a check and show up in GitHub in-line with the files changed tab, just the way lint errors are shown. What do you think?

jpwaves:
> Oops, yes, `src/utils/lib` is definitely generated and should be ignored for prettier. Definitely any generated files should be ignored. What other files are showing need changes? Requesting changes for these include/exclude files stuff.
> 
> I'm not too keen on a pre-commit hook unless you think it's the proper way to do this. I have second-hand bad experience with pre-commit hooks from first co-op which is why I was suggesting trying to get it to run as a check and show up in GitHub in-line with the files changed tab, just the way lint errors are shown. What do you think?

The pre-commit hook I was thinking of was just running the prettier enforce script before the commit to make sure everything is prettified so devs don't need to worry about it, which is what we do at my co-op. I'm not sure what issues you have with pre-commit hooks, but from my experience, they're a nice convenience to have. What exactly went wrong with the pre-commit hooks at your co-op?

jamescd18:
Oh lol if you do it co-op then don’t let me stop you. Go right ahead and lets do it as a pre-commit hook. My bad experience is just the other co-op struggling to ever get a pre-commit hook going for sonarlint scans. Very stupid bad experience, not at all warranting us avoiding it.

jpwaves:
Here's a screenshot of all the files triggering warnings by prettier:
prettier-warnings

jpwaves:
> Oh lol if you do it co-op then don’t let me stop you. Go right ahead and lets do it as a pre-commit hook. My bad experience is just the other co-op struggling to ever get a pre-commit hook going for sonarlint scans. Very stupid bad experience, not at all warranting us avoiding it.

We'll make creating the pre-commit hook a separate ticket then, in case any of the other devs want to try their hands at this.

acd124:
Should we also be checking css files as well? Might be worth including too. I'm more on the fence about json, but it might be considered.

jpwaves:
> Should we also be checking css files as well? Might be worth including too. I'm more on the fence about json, but it might be considered.

Yeah, we should probably check CSS files too. I can convert the json file to  a `.js` file instead - `.json` just happened to be the quickest and easiest way to move the configs.

jpwaves:
New screenshot of tagged files:
prettierlogs1

jamescd18:
Does `.test.tsx` need to be called out separately? Does that mean we also need to call out all our other fun extensions like `.hooks.tsx` or `.api.ts`

jpwaves:
> Does `.test.tsx` need to be called out separately? Does that mean we also need to call out all our other fun extensions like `.hooks.tsx` or `.api.ts`

tested, guess not

jamescd18 (approved):
LGTM!

PR #660, opened 2022-05-25
#658 - Edit Doesnt Update Bullets BUG

### LINKED ISSUE #658: Projects - Edit Creates Change But Doesn't Update WP
### Browser(s)

Chrome, Firefox, Safari

### Platform(s)

Desktop

### Version(s)

production

### Observed Behavior

An implemented change is registered, but update to the actual bullet in the work package is not happening.

### Expected Behavior

Both happen.

### Steps to Reproduce

1. Edit a work package
2. Change a bullet
3. Save
4. Observe change created but bullet does not get altered

### Screenshots (as needed)

_No response_

anthonybernardi:
## Changes

Took out the prisma.$transactions which may be having a weird behavior on prod which is why we can't replicate it.

## Notes

No proof that this is the real solution but it's hard to know without being able to replicate it

## Test Cases

editing bullets still works for me on local after this change

Closes #658

### DIFF
--- a/src/backend/functions/projects-edit.ts
+++ b/src/backend/functions/projects-edit.ts
@@ -340,21 +340,14 @@ const addDescriptionBullets = async (
 
 // edit descrption bullets in the db for each id and detail pair
 const editDescriptionBullets = async (editedIdsAndDetails: { id: number; detail: string }[]) => {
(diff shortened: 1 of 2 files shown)


jamescd18 (approved):
Bet let’s give it a shot

PR #664, opened 2022-05-26
#452 - Loading Spinner

### LINKED ISSUE #452: Auth - Add Loading Spinner to Login Page
## Description

 Add a loading indicator on the page right after the user selects their northeastern email account in the popup window

## Background 

It's a little annoying to sit there on the login page while you wait for the authentication server to respond before it will move you over to the actual application, so as a user, it would be much more informative to see evidence of loading when waiting on the login. Currently, it redirects the user back to the login page where it sits for a good 2-3 seconds before finally letting the user into the application. 

## User Story

As a project manager, I want to see if the dashboard is loading after login, so that I know that the dashboard is simply loading rather than freezing.

## Proposed Solution
- [x] Display a loading indicator on the page as soon as the user selects their account in the popup window

Ari-S-123:
## Changes

Set up a route for just the spinning loading icon, and made it the active component for the portion of time between a successful Google sign-in and the main app loading onto the screen.

## Notes

These changes cannot be tested locally as the Google log-in depends on environment variables only available on the production branch. 

## Checklist

- [x] No linting errors
- [x] No merge conflicts
- [x] All checks passing
- [ ] Screenshots of UI changes (if applicable)

Closes #452


src/services/auth.hooks.ts
```diff
@@ -29,11 +30,17 @@ export const useProvideAuth = () => {
     setUser(undefined);
   };
 
+  const updateLoadingState = () => {
+    setIsLoading(serverSignin.isLoading);
```
jamescd18, line 34: You’re pulling out the correct `isLoading` variable, but I have to ask: where is the `isLoading` useState value you added being used? Your current implementation relies on the return value from calling this function. But when will the code be run again in order to get a different value from the initial call?

src/services/auth.hooks.ts
```diff
@@ -29,11 +30,17 @@ export const useProvideAuth = () => {
...
+    return isLoading;
+  };
+
   return {
     user,
     devSignin,
     signin,
-    signout
+    signout,
+    updateLoadingState
```
jamescd18, line 43: Why expose an update function instead of directly exposing the `isLoading` field?

jamescd18 (changes requested):
This is a good start, but not quite aligned with the hooks loading pattern we’ve used that rides off of react-query loading. I’ve detailed specific line items where I think it would be good to adjust things. I would also strongly recommend taking a look at how we do loading on other pages if you haven’t yet. This is a good example: https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/bf0238ce00de8910a48569199a3e39207a308c1f/src/frontend/pages/CreateWorkPackagePage/create-wp-form.tsx#L43

You want to think about how the react component code is run again on page re-render, and data inside hooks is what might change between renders. Let me know if you need more pointers

jamescd18 (approved):
This is perfect! I tested it locally with the GoogleAuth credentials and it worked flawlessly! I love it! I also really love seeing the journey and how you've learned about mocking since mocking is definitely an advanced concept in testing. It took me a while to understand and I'm glad to see more devs getting it. Great job!

PR #675, opened 2022-05-27
#388 - Saas Bootstrap

### LINKED ISSUE #388: General - Implement custom bootstrap elements using Sass

anthonybernardi:
## Changes

Changed to custom.scss file. Got rid of some !important by chanigng the secondary bootstrap color to be the NER Red.

## Notes

I only did some of this so we can discuss the design approach. Basically Bootstrap seems to encourage using their "variants" for things. The primary variant is the normal blue and if I change that then so many highlights become red which to me looks like errors. My thought process was to then make the secondary variant the NER Red and then any button can just have variant="secondary" and it will be the right color and stuff like that.

I changed some css files to scss but I'm actually realizing I don't need to do that since the custom.scss changing the secondary color does everything for us (at least for what we want right now, scss files will definitely be useful in teh future for functions and stuff).

I'm interested in review on this before I do the whole thing but I think this simplifies a lot.

## Screenshots

Screen Shot 2022-05-27 at 2 19 44 PM
buttons look the same now but i'm just using variant="secondary" on them and $secondary: the NER red color in the custom.scss file.

## To Do

Changing more buttons to use the secondary variant then getting rid of all !important? Will depend on approach decided

Closes #388


src/custom.scss
```diff
@@ -3,12 +3,42 @@
...
 
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
+    'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+code {
```
jamescd18, line 14: is this needed? what's this for?

src/custom.scss
```diff
@@ -3,12 +3,42 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
```
jamescd18, line 8: might be good to have NER fonts in here?

src/custom.scss
```diff
@@ -3,12 +3,42 @@
...
+  -moz-osx-font-smoothing: grayscale;
+}
+
+code {
+  font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New', monospace;
+}
+
+$secondary: #ef4345;
+
+.item-box {
```
jamescd18, line 20: I thought this was replaced by the use of the `PageBlock` component, but you might need to check on that

src/custom.scss
```diff
@@ -3,12 +3,42 @@
...
+
 // change the theme
 $theme-colors: (
-    'ner-red': #ef4345,
-    'ner-dark-grey': #292929,
-    'ner-white': #ffffff
+  'ner-red': #ef4345,
+  'ner-dark-grey': #292929,
+  'ner-white': #ffffff,
+  'ligma': #238423
```
jamescd18, line 36: LIGMA

src/frontend/pages/ChangeRequestsPage/change-requests-filter/change-requests-filter.tsx
```diff
@@ -6,7 +6,7 @@
 import { useState } from 'react';
 import { Button, Card, Dropdown, Form } from 'react-bootstrap';
 import { ChangeRequestType, ChangeRequestReason } from 'utils';
-import styles from './change-requests-filter.module.css';
+import styles from './_change-requests-filter.module.scss';
```
jamescd18, line 9: why the underscore?

jamescd18 (changes requested):
Looking pretty good, just smaller things!

src/custom.scss
anthonybernardi: lolllll

src/custom.scss
anthonybernardi: idk I just copied everything from index.css into here

src/custom.scss
anthonybernardi: I just copied this from the index.css file, what are the NER fonts?

src/custom.scss
anthonybernardi: i think you might be right? taking it out doesnt seem to change anything

src/frontend/pages/ChangeRequestsPage/change-requests-filter/change-requests-filter.tsx
anthonybernardi: I did it because it's general Saas naming ocnvention but turns out it's only for "partial" files which is not really what we're using so I'll get rid of it

src/custom.scss
anthonybernardi: getting rid of it broke nothing so 🤷‍♂️

jamescd18 (approved):

PR #685, opened 2022-05-29
#684 - WP Creation Failing Bug

### LINKED ISSUE #684: [WP Creation Form] - Errors with valid data
### Browser(s)

Chrome, Firefox, Safari

### Platform(s)

Desktop

### Version(s)

main, production

### Observed Behavior

work package form is giving an error with valid data

### Expected Behavior

it should work and make the work package

### Steps to Reproduce

1. fill out the work package form like above
2. press submit

### Screenshots (as needed)

image

anthonybernardi:
## Changes

took out a regex that isn't working, might have to do with the font change? not sure

## Test Cases

- manually tried to make the same WP i was making before i made teh change and now it works

Closes #684

### DIFF
--- a/src/frontend/pages/CreateWorkPackagePage/create-wp-form.tsx
+++ b/src/frontend/pages/CreateWorkPackagePage/create-wp-form.tsx
@@ -142,9 +142,9 @@ const CreateWPForm: React.FC = () => {
         deliverables
       });
       history.push(routes.CHANGE_REQUESTS);
-    } catch (e) {
+    } catch (e: any) {
       console.log(e);
-      alert('something went wrong');
+      alert(e.message);
     }
 
     /**
--- a/src/utils/src/validate-wbs.ts
+++ b/src/utils/src/validate-wbs.ts
@@ -11,13 +11,11 @@ import { WbsNumber } from './types/project-types';
  * @param wbsNum WBS number to validate
  */
 export const validateWBS = (wbsNum: string): WbsNumber => {
-  const errorMsg: string = 'WBS Invalid: ';
+  const errorMsg: string = `WBS "${wbsNum}" is Invalid: `;
   if (wbsNum == null || wbsNum === undefined) {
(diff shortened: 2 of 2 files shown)


src/utils/src/validate-wbs.ts
```diff
@@ -11,13 +11,11 @@ import { WbsNumber } from './types/project-types';
  * @param wbsNum WBS number to validate
  */
 export const validateWBS = (wbsNum: string): WbsNumber => {
-  const errorMsg: string = 'WBS Invalid: ';
+  const errorMsg: string = `WBS "${wbsNum}" is Invalid: `;
```
jamescd18, line 14: this is good, better error messaging is always good

src/utils/src/validate-wbs.ts
```diff
@@ -11,13 +11,11 @@ import { WbsNumber } from './types/project-types';
...
   if (wbsNum == null || wbsNum === undefined) {
     throw new Error(errorMsg + 'given WBS # is null');
   }
-  if (wbsNum.match(/\./g) == null) {
```
jamescd18, line 18: I'm not convinced this is a good thing to remove, and it would also require updating tests

src/frontend/pages/CreateWorkPackagePage/create-wp-form.tsx
```diff
@@ -142,9 +142,9 @@ const CreateWPForm: React.FC = () => {
         deliverables
       });
       history.push(routes.CHANGE_REQUESTS);
-    } catch (e) {
+    } catch (e: any) {
       console.log(e);
-      alert('something went wrong');
```
jamescd18, line 147: This is good, better error messaging

jamescd18 (changes requested):
Left some feedback, I don't think changing the wbs validation function is a good idea

jamescd18:
Unreproducible

PR #694, opened 2022-06-01
#94 themes

### LINKED ISSUE #94: General - Add dark theme
it too bright

anthonybernardi:
## Changes

Added in a theme type, hook, and context (heavily based on James' earlier PR for this) for the frontend. Made a "themes" list with values for things. This keeps it expandable if we want to add more themes (although for now it's just a light vs dark button so that would need to become a dropdown and stuff like that but it wouldn't be too bad)

## Notes

I think the design of this is pretty good but I'm open to ideas on how to improve it. There are also some slightly unfinished aspects to this.

- One thing is that I don't know how to change the background color of the form inputs and I'm not sure if you can at all. 

- Another thing is that I don't know how to store a user's theme choice (without making it part of the backend which we could definitely do). This means that either light or dark has to be the default and it will get reset to the default every time the page is refreshed. Right now I put dark as the default.

Aside from that this feature is pretty finished and could probably be shipped if people wanted it really badly and were fine with these problems.

## Test Cases

not sure what to test since it's all just colors? is that something you can test? I've looked through every page and manually it all works

Here's a video of a few of the pages (this was before I changed the button which now goes from Night to Light):

https://user-images.githubusercontent.com/63663597/171473627-0354e2e3-a1f8-4ad1-b489-cc09b7df0654.mp4

## To Do

- [ ] consult with design team about colors
- [ ] test things?
- [ ] find out how to set background color of form inputs
- [ ] find out best way to save user's preference of theme
- [ ] fix the broken tests

Closes #94


src/services/theme.hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ThemeContext } from '../frontend/app/app-context-theme/app-context-theme';
+import { Theme } from '../shared/types';
+import themes from '../shared/themes';
+
+// Provider hook that creates theme object and handles state
+export const useProvideTheme = () => {
+  // eslint-disable-next-line prefer-destructuring
+  const defaultTheme = themes[0];
+
+  const [bgColor, setBgColor] = useState(defaultTheme.bgColor);
```
jamescd18, line 16: Why have individual useStates when you could just have 1 for the Theme object?

src/services/theme.hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { useState, useContext } from 'react';
+import { ThemeContext } from '../frontend/app/app-context-theme/app-context-theme';
+import { Theme } from '../shared/types';
+import themes from '../shared/themes';
+
+// Provider hook that creates theme object and handles state
+export const useProvideTheme = () => {
+  // eslint-disable-next-line prefer-destructuring
+  const defaultTheme = themes[0];
```
jamescd18, line 14: Why have this when you could just instantiate the useState with whatever the default theme is?

src/shared/types.ts
```diff
@@ -11,3 +11,11 @@ export interface Auth {
   signin: (token: string) => Promise<User>;
   signout: () => void;
 }
+
+export interface Theme {
+  toggleTheme?: (theme: Theme) => void;
```
jamescd18, line 16: This part doesn't feel quite right. The hook probably shouldn't be returning the same interface that defines what a theme is.

src/frontend/pages/ProjectsPage/projects-table-filter/projects-table-filter.tsx
```diff
@@ -146,7 +148,7 @@ const ProjectsTableFilter: React.FC<FilterProps> = ({ onClick, leads, managers }
 
   return (
     <>
-      <Card>
+      <Card bg={theme.cardBg} className={theme.themeName}>
```
jamescd18, line 151: Do you think maybe we could have an abstracted card component the same way we have the page block component? Ideally this would significantly reduce the number of locations we're using the theme hook

src/frontend/layouts/nav-top-bar/nav-top-bar.tsx
```diff
@@ -29,6 +42,7 @@ const NavTopBar: React.FC = () => {
       <Navbar.Toggle aria-controls="nav-top-bar-items" />
       <Navbar.Collapse id="nav-top-bar-items">
         <Nav className="ml-auto">
+          <Button onClick={swap}>{themeIndex === 0 ? 'Light' : 'Night'}</Button>
```
jamescd18, line 45: I was actually hoping to have this be located on the settings page, but I'm open to reasons why not if you think it's better elsewhere.

Also, I think you probably could do this as a dropdown and then you could avoid hard-coding the label text and have `themes.map(t => t.themeName)` or something

src/shared/themes.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import { Theme } from './types';
+
+const themes: Theme[] = [
+  {
+    themeName: 'dark',
```
jamescd18, line 10: why not just `name`?

src/shared/themes.ts
```diff
@@ -0,0 +1,23 @@
...
+    bgColor: '#353434',
+    cardBg: 'dark',
```
jamescd18, line 12: These aren't really specific to cards tho, right? It's just that with bootstrap we can define an background and border and swap which one is color A and which is color B, right?

src/custom.scss
```diff
@@ -16,22 +16,45 @@ body {
...
+  'ner-dark-bg': #353434,
+  'ner-light-bg': #ffffff,
   'ner-white': #ffffff
 );
 
 @function theme-color($key) {
   @return map-get($theme-colors, $key);
 }
 
+.dark {
```
jamescd18, line 43: feels funny for some elements of theme to be stored in the theme object and then some other ones in here...?

src/frontend/app/app-public/app-public.tsx
```diff
@@ -13,29 +14,36 @@ import './app-public.module.css';
 const AppPublic: React.FC = () => {
   const auth = useAuth();
   const history = useHistory();
+  const theme = useTheme();
+
+  // eslint-disable-next-line prefer-destructuring
+  document.body.style.backgroundColor = theme.bgColor;
```
jamescd18, line 20: Very uhhhhhhhh not good?

jamescd18 (changes requested):
This is great! Much better than my attempt which I quickly gave up on. Conceptually speaking, I think my major points are:
- I initially thought I would build theme into a broader settings context and hook, but I'm definitely open to theme having its own dedicated context and hook if you'd prefer
- I wanted to hide this away in the setting page, similar to how Slack does it. I know some sites like Mozilla MDN Web Docs have the dark/light slider at the top on every page, but we're a web app not a website so I think it makes more sense to be in settings
- it could be a slider if you use `<Form.Switch>` or it could be a dropdown if you want it inherently expandable beyond 2 options
- I would encourage pushing all the logic of switching into the hook so the hook only exposes the current theme and a swap function with signature of `() => void`
- Back-end we can absolutely add stuff. I've been thinking we should have User Profiles to store things like major, grad year, etc. We could choose to put settings in user profile or we could have settings be separate. I've also been thinking I want organization settings, but that's a separate topic. We should probably just have a new back-end endpoint for this. Then we would need a hook and api func for mutating. I can build this if you want

src/services/theme.hooks.ts
anthonybernardi: true

src/services/theme.hooks.ts
anthonybernardi: true

src/shared/types.ts
anthonybernardi: true

src/frontend/pages/ProjectsPage/projects-table-filter/projects-table-filter.tsx
anthonybernardi: i think that would be a good idea for a future ticket. I honestly don't know how similar all their construction is though

src/frontend/layouts/nav-top-bar/nav-top-bar.tsx
anthonybernardi: i think as of right now i want it on the main page because of the whole refresh cancels your selection thing so I don't want users to have to go back to settings every time they want to change it back. once we set it up as being part of the settings so that it's persistent and stuff I agree

src/shared/themes.ts
anthonybernardi: tru

src/shared/themes.ts
anthonybernardi: I just set it up like this to kinda separate everything we're using in case we start doing stuff on a different component and we want to use other variants. also not every component has the bg= and border= properties

src/custom.scss
```diff
@@ -16,22 +16,45 @@ body {
...
+h4,
+h5 {
   font-family: $ner-header-font;
 }
 
 $primary: #ef4345;
+$dark: #464646;
 
 // change the theme
 $theme-colors: (
```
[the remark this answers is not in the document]
anthonybernardi: this defines what our dark variant color is for bootstrap (so every card bg that has bg='dark' is using this color)

src/custom.scss
anthonybernardi: it is kinda funny but this is what the themeName (now just name) field is referring to. i don't think there's a way to define js values like "dark" and "light" for variants as well as css classes all in one spot if that makes sense, so that's why it's separate but the themes do point to these class names in a sense

src/frontend/app/app-public/app-public.tsx
```diff
@@ -13,29 +14,36 @@ import './app-public.module.css';
...
+
   return (
-    <Switch>
-      <Route path={routes.LOGIN}>
-        <Login
-          postLoginRedirect={{ url: history.location.pathname, search: history.location.search }}
+    <html className={theme.themeName}>
```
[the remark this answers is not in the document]
anthonybernardi: i agree that it's weird af but it's the only way i could figure out to apply the classes in the custom.scss file universally without doing a ton of work

src/frontend/layouts/nav-top-bar/nav-top-bar.tsx
```diff
@@ -3,17 +3,30 @@
...
+import { useTheme } from '../../../services/theme.hooks';
+import themes from '../../../shared/themes';
 
 const NavTopBar: React.FC = () => {
   const auth = useAuth();
+  const theme = useTheme();
+
+  const themeIndex = themes.findIndex((element) => element.themeName === theme.themeName);
+
+  const swap = () => {
```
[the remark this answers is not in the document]
anthonybernardi: true!

anthonybernardi:
To address your bullet points in your review @jamescd18:

* I do agree that we should move the toggle to settings but I think we should wait to do that until it gets saved. Otherwise they would have to go to settings every time they want to change it after any page refresh.
* The logic was really always in the hook of actually switching, it was just about which one you wanted to switch to that you had to decide first which was what I was doing with the index stuff. I have it as a (name:string) => void so that it is expandable beyond just dark/light
* saving it in the backend sounds good, we can just have it save the name of the theme or something, although in that case I guess it might be part of a settings object so maybe this should be too

jamescd18 (changes requested):

src/frontend/components/description-list/description-list.test.tsx
```diff
@@ -4,10 +4,26 @@
...
+import { useTheme } from '../../../services/theme.hooks';
+import { Theme } from '../../../shared/types';
 import { exampleWorkPackage2 } from '../../../test-support/test-data/work-packages.stub';
 import DescriptionList from './description-list';
 
+jest.mock('../../../services/theme.hooks');
+const mockTheme = useTheme as jest.Mock<Theme>;
+
+const mockHook = () => {
+  mockTheme.mockReturnValue({
```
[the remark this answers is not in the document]
anthonybernardi: true

jamescd18 (approved):

Issue #666, opened 2022-05-26
Projects - Remove Spread Operator from `projects.ts` Back-End Endpoint


jamescd18:
## Overview

Ensure the `projects.ts` endpoint is only providing the data required for the request and not exposing extra fields that are not needed.

### Description Bullet Converter

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/backend/functions/projects.ts#L96-L100

The object returned by this function should be limited to the following fields:
https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/utils/src/types/project-types.ts#L63-L68

### Project

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/backend/functions/projects.ts#L112-L131

The object returned by the project transformer should be limited to the following fields:

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/utils/src/types/project-types.ts#L18-L27

and these

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/utils/src/types/project-types.ts#L35-L48

### Change

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/backend/functions/projects.ts#L119-L122

The implemented change object should be limited to the following fields:

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/utils/src/types/change-request-types.ts#L79-L86

### Work Package

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/backend/functions/projects.ts#L132-L157

The object should be limited to the following fields

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/3f6a69d3eaa925c727d1c82fbca216040b12a8db/src/utils/src/types/project-types.ts#L18-L27

and these

[...]

PR #747, opened 2022-06-23
#666 - Removed the spread operator wherever it appeared in the code

tigerhippo:
## Changes

I removed the spread operator wherever it appeared in the code. 

## Notes

I am not sure what to input for the fields projectLead and projectManager on lines 155 and 156 respectively. I did workPackage.wbsElement.projectLead and workPackage.wbsElement.projectManager (which should work because the WbsElement interface had all of these fields). However, the error message stated that the wbsElement did not have any of these fields.

I still need to test my code, but it would be great if someone could take a look at all the places where I removed the spread operator to see if I did it correctly!


src/backend/functions/projects.ts
```diff
@@ -138,21 +145,33 @@ const projectTransformer = (
...
-        ...workPackage.wbsElement,
+        //...workPackage,
+        //...workPackage.wbsElement,
         id: workPackage.workPackageId,
         wbsNum: wbsNumOf(workPackage.wbsElement),
+        dateCreated: workPackage.wbsElement.dateCreated,
+        name: workPackage.wbsElement.name,
         status: convertStatus(workPackage.wbsElement.status),
+        projectLead: workPackage.wbsElement.projectLead,
+        projectManager: null,
```
jamescd18, line 156:
```suggestion
        projectManager: workPackage.wbsElement.projectManager,
```

src/backend/functions/projects.ts
```diff
@@ -138,21 +145,33 @@ const projectTransformer = (
       );
 
       return {
-        ...workPackage,
-        ...workPackage.wbsElement,
+        //...workPackage,
+        //...workPackage.wbsElement,
```
jamescd18, line 149:
```suggestion
```

jamescd18 (changes requested):
Looking good! Give those a try and attach screenshots of errors if you run into any. You can also still commit + push your code and see if the automated test runner will fail or not. If it doesn't, it's probably still good code.

anthonybernardi (changes requested):
great work! the only thing I want to add from James' review is that you committed the package-lock.json file which should not be committed except in very specific circumstances.

To uncommit this, you can essentially commit the current "main" version of the package-lock to your branch. This way, the package-lock will be the same as the one on main and then it wont be a change if that makes sense

anthonybernardi:
i swear the package-lock is the exact same as the one on main. Go check line ~1496 for proof that the diff is wrong

anthonybernardi:
image

this makes no sense and it's the same as on main so im merging this

anthonybernardi (approved):

Issue #730, opened 2022-06-17
Users - Enable Users to Save Settings


jamescd18:
### Current Features

User settings are entirely client side. Nothing is saved to the database.

### Desired Additional Features

Enable users to save their settings so that it will load in with settings when they log in from another device or during another session.

PR #757, opened 2022-06-30
#730 add user settings

jamescd18:
## Changes

Added user settings to the database model and prisma and seed data.

Okay so basically my plan here was to separate out and have the login endpoint be what first returns the user settings, since this would avoid an extra call to the server and allow us to set up the app with the user's preferred settings as soon as they log in.

This also goes along with my plan to change around the user endpoints to remove sending unnecessary info to the front-end whenever possible. Having a singular unified User interface makes that somewhat difficult, so this take a first step to try and have a separate interface for authenticated user and see how things go.

## Screenshots

Screen Shot 2022-07-02 at 3 42 11 AM

Screen Shot 2022-07-02 at 3 42 28 AM

Screen Shot 2022-07-02 at 3 42 52 AM

instant change post-change
Screen Shot 2022-07-02 at 3 43 17 AM

preserved on reload, theme is loaded on login
Screen Shot 2022-07-02 at 3 43 41 AM

## Checklist

- [x] No linting errors
- [x] No merge conflicts
- [x] All checks passing


src/shared/types.ts
```diff
@@ -3,7 +3,7 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
-import { User } from '@prisma/client';
+import { User } from 'utils';
```
anthonybernardi, line 6: good catch!

anthonybernardi: on all of these didn't realize we did this so much

src/backend/prisma/schema.prisma
```diff
@@ -53,6 +58,9 @@ model User {
   email                   String           @unique
   emailId                 String?          @unique
   role                    Role             @default(GUEST)
+  userSettingsId          String?          @unique
+  userSettings            UserSettings?    @relation(fields: [userSettingsId], references: [id])
```
anthonybernardi, line 62: I think that the user should have just the userSettings object and the userSettings model should be concerned with the id and @relation tag. I don't know if it matterrs but it makes more rsense to me in the sense that

anthonybernardi, line 62: I think that the user should have just the userSettings object and the userSettings model should be concerned with the id and @relation tag. I don't know if it matters but it makes more sense to me in the sense that the user settings is then tied to the user

anthonybernardi (commented):
i like it. so is the plan then to basically make the authenticated user thing part of a context when logged in so it gets stored and stuff so we only need it once?

src/shared/types.ts
jamescd18: Lmao me neither!

src/backend/prisma/schema.prisma
jamescd18: Lemme experiment and I'll see if it's possible, I can't remember what did / didn't work

jamescd18:
Yeah basically. I guess it kinda depends on how we expand user settings and what we store in there. We would separately enable editing of user settings via the settings page.

src/backend/prisma/schema.prisma
jamescd18: Like this?

Screen Shot 2022-07-02 at 3 40 26 AM

anthonybernardi: yeee

src/frontend/app/app-public/app-public.tsx
```diff
@@ -20,7 +20,7 @@ const AppPublic: React.FC = () => {
   document.body.style.backgroundColor = theme.bgColor;
 
   return (
-    <html className={theme.name}>
+    <html className={theme.cardBg}>
```
anthonybernardi, line 23: this is not always true in the case of expanded themes -- the name field of the theme is meant to be the class name

anthonybernardi (changes requested):
This is very cool. Here are some initial notes, haven’t gone fully in depth but I’ve looked at everything so far:
- Will want to get rid of the top theme picker
- Thomas emrax’s theme doesn’t load in as light on my local even if it’s set as his default theme
- I think that all users will naturally have no settings because it’s optional and then all the endpoints will not work because they assume it already exists, when they should probably create a default settings if it doesnt exist

src/frontend/app/app-public/app-public.tsx
jamescd18: we'll need some kind of translation from a given theme to a unique identifier. what do you suggest?

jamescd18:
> Will want to get rid of the top theme picker

I figured we could keep it. did you think it's better removed? cuz if we remove it then there's technically no way to change your current theme without changing your default theme, right? Is this what you'd want?

> Thomas emrax’s theme doesn’t load in as light on my local even if it’s set as his default theme

hmmm, any ideas why?

> I think that all users will naturally have no settings because it’s optional and then all the endpoints will not work because they assume it already exists, when they should probably create a default settings if it doesnt exist

I was figuring we would fix this by manually generating settings for everyone when we deploy this, which I think is no biggie given the manual script

src/frontend/app/app-public/app-public.tsx
anthonybernardi: and then className={theme.className}

anthonybernardi:
ok so I see why it doesn't work on dev signing and im pretttty sure this will work on prod. The one commit I made was just so that it uses the chosen theme instead of the default theme value in the hook

looks good to me!

src/backend/functions/users.ts
```diff
@@ -97,7 +124,38 @@ const logUserIn: ApiRouteFunction = async (_params, event) => {
...
+      where: { userId },
+      data: { userSettings: { create: {} } },
+      include: { userSettings: true }
+    });
+  }
+  return buildSuccessResponse(user.userSettings!);
+};
+
+/** Update settings for the specified user */
+const updateUserSettings: ApiRouteFunction = async (params: { id: string }, event) => {
```
anthonybernardi, line 149: what if this user doesn't have settings yet? I guess we're relying on the user going to the settings page first which should call "getUserSettings" for them and create the settings object

anthonybernardi (commented):
our auth system is kinda elaborate

src/backend/functions/users.ts
jamescd18: correct. I'm still maybe sorta a fan of just editing the data so there is no chance for a user to ever not have a user settings object. ignoring that though, the get request will create it. and the update form can't be accessed without calling the get request, so it's guaranteed to always exist for the update. the only scenario possible is if someone tries to update via API access. we could change to upsert. we would still need an answer for what to return if get has none though. lmao the more you know. halfway through writing this I googled it and: https://github.com/prisma/docs/issues/640

so basically either we
- edit data and assume it's always there
- leave it as is and say meh
- change both to upsert cuz that's most robust

jamescd18: > change both to upsert cuz that's most robust

I picked this one cuz it's more robust and actually less code I think

src/services/theme.hooks.ts
```diff
@@ -6,10 +6,13 @@
...
 import { ThemeContext } from '../frontend/app/app-context-theme/app-context-theme';
 import themes from '../shared/themes';
+import { useAuth } from './auth.hooks';
 
 // Provider hook that creates theme object and handles state
 export const useProvideTheme = () => {
-  const [theme, setTheme] = useState(themes[0]);
+  const auth = useAuth();
+  const defaultThemeIndex = themes.findIndex((t) => auth.user?.defaultTheme === t.name);
+  const [theme, setTheme] = useState(themes[defaultThemeIndex > -1 ? defaultThemeIndex : 0]);
```
jamescd18, line 15: Sorta still confused how this helps / what this intended to do? Is this intended to solve the dev user default theme not working?

The way I set it up was
1. visit login page
2. default dark always used via index 0
3. auth success pulls default theme and toggles as-needed
4. auth'd app loads with user's default theme`theme`auth success pulls default theme and toggles as-needed
5. auth'd app loads with user's default theme`default dark used
6. auth success pulls default theme and toggles as-needed
7. auth'd app loads with user's default theme

does this addition suggest that the theme context is rebuilt whenever auth changes? wouldn't that be redundant with toggling theme in login success call? or perhaps at least the remove the login success theme toggle in favor of this?

anthonybernardi (approved):
nice

Issue #554, opened 2022-02-25
Change Requests - Implemented Changes Should Not Say "WBS Element"


jamescd18:
## Description

Implemented Changes should not say "WBS element" at all. This is something we don't need to reference on the application as this word means nothing for users. We should not see "WBS element" at all in the application, but instead what the WBS element is referring to.

## Background
Currently, under the implemented changes of a change request, we have "WBS element" being referenced on the application. Users are likely not going to be familiar with what a "WBS Element" is and this change would also prevent us from exposing behind-the-scenes details of the application to the users. This issue is caused by the implementation of the work package edit endpoint, specifically in https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/8d37ab9191269718eb8e6d3823f10ac701015917/src/backend/functions/work-packages-edit.ts#L105-L106

## Proposed Solution

Edit the Implement change details of a Change Request to replace the "WBS element" with its correct model, either "Project", "Work Package", or "Change Request". 

## Screenshots

Example of "WBS Element" where we would like to replace it with "Work Package". It would be better if it said `Edited Work Package status from "INACTIVE" to "ACTIVE"`. 

<img width="561" alt="Screen Shot 2022-02-25 at 6 46 40 PM" src="https://user-images.githubusercontent.com/51571627/155817833-7db9b572-2cdc-4e0a-8af9-b2c330912698.png">

PR #759, opened 2022-07-07
#554 changes should not say wbs element

tigerhippo:
I changed 'wbs element status' to just 'status' in the work-packages-edit.ts file. 

I was initially going to change 'wbs element status' to 'work package status', because the file relates to work packages, and that would be easier for the user to understand than stating 'wbs element' (which the user likely doesn't know about). However, after discussing with Anthony, I decided to change 'wbs element status' to just 'status', because this would keep everything consistent (none of the other fields have 'wbs element' in front of them, even though they could). It is also pretty obvious to the user what the status is referring to (they know that it is referring to a work package, because they have just created a work package). 

An image of my change is below.

Screen Shot 2022-07-07 at 12 02 20 PM

I have already completed 'npm run test' and ran all test suites on my code, which worked fine. 

It would be great if someone could review my change!


anthonybernardi:
Hm so I think you made this branch off of the branch you were working on for the previous ticket because it still has all of your changes from before. Before making a new branch, you should switch to main ("git checkout main") and then pull ("git pull") to get the latest updates from main. Then you can make a new branch from there so that you're making changes to the current state of the repo.

PR #764, opened 2022-07-18
#709-added-risk-edit-files-with-test

callmejim1226:
## Changes

I added three files (risk-edit.ts, risk-types.ts, and risk-edit.test.ts) to develop the endpoint for editing a risk. In risk-edit.ts, I validated incoming data using middy (riskEditInputSchemaBody from risk-types.ts), edited the selected risk from the database, and finally let this endpoint return the edited risk from the database. 

## Notes
Need to provide screenshots and test the endpoint manually to make sure the endpoint for editing risk works. 

## Test Cases

- Case A: Check if the event object fails (statuscode: 400) when the incoming inputs to the risk edit endpoint have no Risk id.
Result: Pass 
- Case B: Check if the event object fails (statuscode: 400) when the invalid user is trying to edit the Risk from the database. 
Result: Pass

_Any relevant screenshots go here_

## To Do

_Any remaining things that need to get done_

- [x] Fix the error on lines 33 and 42 in risk-edit.ts
- [ ] Check up with the implementation of updating the risk from the database in risk-edit.ts
- [ ] Manually testing

## Checklist

- [x] No linting errors
- [ ] No merge conflicts
- [x] All checks passing
- [ ] Screenshots of UI changes (if applicable)
- [ ] Remove any not-applicable sections

Closes # (issue #)

### DIFF
--- a/src/backend/functions/risk-edit.ts
+++ b/src/backend/functions/risk-edit.ts
@@ -23,37 +23,24 @@ const prisma = new PrismaClient();
 export const editRisk: Handler<FromSchema<typeof inputSchema>> = async ({ body }, _context) => {
   const { userId, id, detail, resolved } = body;
 
-  // verify user is allowed to edit work packages
+  // verify user is allowed to edit risk
   const user = await prisma.user.findUnique({ where: { userId } });
   if (!user) return buildNotFoundResponse('user', `#${userId}`);
   if (user.role === Role.GUEST) return buildNoAuthResponse();
 
   // get the original risk and check if it exists
-  const originalRisk = await prisma.risk.findUnique({
-    where: { id }
-  });
-  if (originalRisk === null) {
-    return buildNotFoundResponse('Risk', `#${id}`);
-  }
+  const originalRisk = await prisma.risk.findUnique({ where: { id } });
+  if (!originalRisk) return buildNotFoundResponse('Risk', `#${id}`);
 
   // update the risk with the input fields
   const updatedRisk = await prisma.risk.update({
-    where: {
-      id
-    },
-    data: {
-      detail,
-      isResolved: resolved
-    }
+    where: { id },
+    data: { detail, isResolved: resolved }
   });
 
   // return the updated risk
-  return buildSuccessResponse(updatedRisk);
+  return buildSuccessResponse({ detail: updatedRisk.detail, isResolved: updatedRisk.isResolved });
 };
-
-/**
- * 	HELPER METHODS:
- */
 
 // expected structure of json body
 const inputSchema = eventSchema(riskEditInputSchemaBody);
--- a/src/backend/functions/risk-edit.ts
+++ b/src/backend/functions/risk-edit.ts
@@ -18,46 +18,38 @@ import {
 } from 'utils';
 import { FromSchema } from 'json-schema-to-ts';
 
-const prisma = new PrismaClient(); 
+const prisma = new PrismaClient();
 
-export const editRisk: Handler<FromSchema<typeof inputSchema>> = async (
-	{ body },
-	_context
-) => {
-	const {
-		userId,
-		id, 
-		detail, 
-		resolved
-	} = body; 
-	
-	// verify user is allowed to edit work packages
-	const user = await prisma.user.findUnique({ where: { userId } });
-	if (!user) return buildNotFoundResponse('user', `#${userId}`);
-	if (user.role === Role.GUEST) return buildNoAuthResponse();
+export const editRisk: Handler<FromSchema<typeof inputSchema>> = async ({ body }, _context) => {
+  const { userId, id, detail, resolved } = body;
 
-	// get the original risk and check if it exists
-	const originalRisk = await prisma.risk.findUnique({
-		where: { id },
-	});
-	if (originalRisk === null) {
-		return buildNotFoundResponse('Risk', `#${id}`);
-	}
+  // verify user is allowed to edit work packages
+  const user = await prisma.user.findUnique({ where: { userId } });
+  if (!user) return buildNotFoundResponse('user', `#${userId}`);
+  if (user.role === Role.GUEST) return buildNoAuthResponse();
 
-	// update the risk with the input fields
-	const updatedRisk = await prisma.risk.update({
-		where: { 
-			id
-		},
-		data: {
-			detail,
-			isResolved: resolved
-		}
-	});
+  // get the original risk and check if it exists
+  const originalRisk = await prisma.risk.findUnique({
+    where: { id }
+  });
+  if (originalRisk === null) {
+    return buildNotFoundResponse('Risk', `#${id}`);
+  }
 
-	// return the updated risk 
-	return buildSuccessResponse(updatedRisk); 
-}
+  // update the risk with the input fields
+  const updatedRisk = await prisma.risk.update({
+    where: {
+      id
+    },
+    data: {
+      detail,
+      isResolved: resolved
+    }
+  });
+
+  // return the updated risk
+  return buildSuccessResponse(updatedRisk);
+};
 
 /**
  * 	HELPER METHODS:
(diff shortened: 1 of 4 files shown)


src/backend/functions/risk-edit.ts
```diff
@@ -0,0 +1,66 @@
...
+  riskEditInputSchemaBody
+} from 'utils';
+import { FromSchema } from 'json-schema-to-ts';
+
+const prisma = new PrismaClient();
+
+export const editRisk: Handler<FromSchema<typeof inputSchema>> = async ({ body }, _context) => {
+  const { userId, id, detail, resolved } = body;
+
+  // verify user is allowed to edit work packages
```
anthonybernardi, line 26: should say risks not work packages

src/backend/functions/risk-edit.ts
```diff
@@ -0,0 +1,66 @@
...
+  const user = await prisma.user.findUnique({ where: { userId } });
+  if (!user) return buildNotFoundResponse('user', `#${userId}`);
+  if (user.role === Role.GUEST) return buildNoAuthResponse();
+
+  // get the original risk and check if it exists
+  const originalRisk = await prisma.risk.findUnique({
+    where: { id }
+  });
+  if (originalRisk === null) {
```
anthonybernardi, line 35: I prefer `if (!risk)` -- you can copy line 28 too where it's all on one line

src/backend/functions/risk-edit.ts
```diff
@@ -0,0 +1,66 @@
...
+      isResolved: resolved
+    }
+  });
+
+  // return the updated risk
+  return buildSuccessResponse(updatedRisk);
+};
+
+/**
+ * 	HELPER METHODS:
```
anthonybernardi, line 55: this can be deleted

src/utils/src/types/risk-types.ts
```diff
@@ -0,0 +1,16 @@
...
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import { FromSchema } from 'json-schema-to-ts';
+import { bodySchema, intType, stringType, booleanType } from './api-utils-types';
+
+export const riskEditInputSchemaBody = bodySchema({
+  userId: intType,
+  id: intType,
```
anthonybernardi, line 11: if you look at the schema.prisma, you can see we actually use a string id for risks (which we should be doing everywhere but woops lol) so this should be stringType. This will also fix the compile errors in risk-edit.ts that you're getting

anthonybernardi (changes requested):
Great stuff! Very close to being done in my opinion, just some very minor changes here.

The only thing left to do after this is to test it manually to make sure it works!

src/backend/functions/risk-edit.ts
```diff
@@ -0,0 +1,66 @@
...
+      id
+    },
+    data: {
+      detail,
+      isResolved: resolved
+    }
+  });
+
+  // return the updated risk
+  return buildSuccessResponse(updatedRisk);
```
jamescd18, line 51: Maybe need transformer?

src/backend/functions/risk-edit.ts
```diff
@@ -0,0 +1,66 @@
...
+  const originalRisk = await prisma.risk.findUnique({
+    where: { id }
+  });
+  if (originalRisk === null) {
+    return buildNotFoundResponse('Risk', `#${id}`);
+  }
+
+  // update the risk with the input fields
+  const updatedRisk = await prisma.risk.update({
+    where: {
```
jamescd18, line 41: Prettier formatting

src/backend/tests/risk-edit.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+describe('work package edit', () => {
+  describe('handler', () => {
+    const func = async (event: any) => {
+      return handler(event, mockContext, () => {});
+    };
+
+    describe('validate inputs', () => {
+      const goodBody = {
+        id: 1,
```
jamescd18, line 17: IDs strings, best to use good example like `ef14de-32bc7`

src/backend/tests/risk-edit.test.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import { mockContext } from '../../test-support/test-data/test-utils.stub';
+import { handler } from '../functions/risk-edit';
+
+describe('work package edit', () => {
```
jamescd18, line 9: Fix test suite name

jamescd18 (changes requested):
Great start! Left comments

src/backend/functions/risk-edit.ts
callmejim1226: Can you elaborate more on the transformer? Like I compared with other editing endpoints, and for project-edit endpoint, I saw that it also just returned buildsuccessresponse at the end.

src/backend/tests/risk-edit.test.ts
callmejim1226: I was not sure how to access that example, but I just used '1' instead of 1. Would that be enough?

src/backend/functions/risk-edit.ts
jamescd18: Yes, if I recall correctly, the response from those endpoints isn't used by any of the front-end, so it doesn't matter how the response object is formed. But if we want to do it properly and use the response object, we should ensure that it is properly formatted. 

For the get endpoints, you should be able to observe that a transformer function, typically with several helper functions, is used to define the proper output object. 

One such example is that the database object may include a WBS_Element ID, but this is typically not something we want to provide to the front-end, so in the transformer function, we do not include it as a field in the return object. 

Another good example is WBS number. In the database, it is stored as 3 fields with a combined unique constraint. In the transformer functions, we typically have a helper function that takes in a WBS_Element database object and returns a proper WBS Number object for the front-end to use. Notably, the front-end expected WBS number to be an object containing the 3 fields, rather than just the 3 fields.

src/backend/tests/risk-edit.test.ts
jamescd18: Yes, but I think copy and pasting my example ID would probably be better to reduce confusion in the future. People might be confused as to why 1 is a string instead of a number. The reality of our database is that the string '1' would not be an actual realistic ID.

src/backend/functions/risk-edit.ts
jamescd18: Try putting each clause (where, data) on one line each.

```
 where: { id },
 data: { id, detail }
```

I believe both are valid for prettier, but one is fewer lines of code and thus more compact.

src/backend/tests/risk-edit.test.ts
anthonybernardi: Even something like "abc" is better just to avoid people thinking it's a number without seeing the quotation marks

src/backend/functions/risk-edit.ts
```diff
@@ -23,38 +23,25 @@ const prisma = new PrismaClient();
...
-      detail,
-      isResolved: resolved
-    }
+    where: { id },
+    data: { detail, isResolved: resolved }
   });
 
   // return the updated risk
-  return buildSuccessResponse(updatedRisk);
+  return buildSuccessResponse({ detail: updatedRisk.detail, isResolved: updatedRisk.isResolved });
```
callmejim1226, line 42: I was not 100% sure if this is what James wanted me to do. I referred to #709 ticket and thought that we would only need a field of detail and isResolved for updated risk in the front-end. Correct me if I'm wrong with this transformer.

jamescd18: a transformer would be a function, but that's a good start. without further tweaks to the front-end though, I don't think the response would be used for anything

anthonybernardi: yeah the transformer is actually unnecessary here since this isn't a GET route but that's the right idea

PR #769, opened 2022-07-21
#696 create risk log visuals

### LINKED ISSUE #696: Projects - Create Risk Log Visuals
### Description

This ticket is to build the frontend components associated with the new risk log feature that is requested by project managers. You only need to make the front end components — another ticket will handle collecting and transmitting the data to and from the endpoint.

### Background

- Each risk will be shown as an item in a checklist
- Each **unresolved** risk will have a 'Convert to CR' button on the right-hand side of the risk. There should not be a delete button
- Each **resolved** risk will have its details crossed-out and a delete button on the right-hand side of the risk. There should not be a 'Convert to CR' button
- Users should be able to resolve and unresolve a risk as many times as they wish. The buttons and state of the risk should change dynamically as users do so
- There will always be a text box present at the end of the list with an add button on the right-hand side. This is so that users can add risks at any time.

### Proposed Solution

- [x] Create a new component "CheckList" that is similar to the "BulletList" component. However, rather than rendering the list items as an ordered or unordered list, render the list items as a checklist.
- [x] Create a new folder "risk-log" in the [project-view-container](https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/tree/main/src/frontend/pages/ProjectDetailPage/project-view-container) folder that houses files for the logic outlined below.
- [x] For each **unresolved** risk, there should be a convert to CR button on the right-hand side of each risk.
- [x]  For each **resolved** risk, the details of the risk should be crossed out and there should be a delete button on the right-hand side of the risk. The delete button should only be present if a risk is resolved.
- [x] At the end of the list, there should be a text box with an add button on the right hand side of the text box. This text box and button should always be present, whether or not there are any list items present. There should be placeholder text within the textbox that reads "Add a new risk".

### Mocks
Image from iOS

z-roth:
## Changes

- Added UI components for risk log feature (display risks, delete completed risk, create change request for risk, add new risk)

## Notes

- One of the tests that checks that the Convert to CR button renders doesn't work for some strange reason, but commented it out
- Doesn't like to work with dark mode, not sure how to implement that yet

### CheckList
- Renders pageblock with given title and header right 
- Renders correct buttons for both resolved and unresolved risks (delete, convert to CR)
- Checkboxes that are marked as checked in the data are checked

### RiskLog
- Renders the Risk Log title
- Renders all risks

image

## Checklist

- [x] No linting errors
- [x] No merge conflicts
- [x] All checks passing

Closes #696


src/frontend/pages/ProjectDetailPage/project-view-container/project-view-container.tsx
```diff
@@ -15,9 +15,11 @@ import PageTitle from '../../../layouts/page-title/page-title';
 import PageBlock from '../../../layouts/page-block/page-block';
 import ProjectDetails from './project-details/project-details';
 import RulesList from './rules-list/rules-list';
+import RiskLog from './risk-log/risk-log';
 import './project-view-container.module.css';
 import { routes } from '../../../../shared/routes';
 import ProjectGantt from './project-gantt/project-gantt';
+import { truncateSync } from 'fs';
```
anthonybernardi, line 22: pls delete

anthonybernardi:
The code looks good but I'll do a full code review when everything is more finalized. As far as design goes, I'd prefer:
1. each risk row to be thinner. The whole thing is very large rn
2. i also might want a way to make adding them harder. not sure what the best design for that would be
3. for dark mode, you're gonna have to mess with the listgroupitem or something idk

jamescd18:
Making it harder to create could be as simple as putting it in a modal. 

Also you might wanna do away with dividing lines and background on each Risk. Different bootstrap component (or just a divs) might help with this. 

Also if you wanna get fancier with UI, you make the right side buttons be icons instead, only show on hover, and maybe add tooltips to say their function after mouse hovers for 2 seconds. This is very justifiable because rn as buttons, you're calling more attention to them than they really deserve. The number of primary buttons on a page at a given time should be kept meaningfully minimal.

z-roth:
Changed around a few things:
image
image

- Removed ListGroup and ListGroupItem and replaced with divs for cleaner look, also works fine now with dark mode
- Removed input text box and added pop-up modal that prompts user for risk input

Going to look into how to add some icons and tooltips.

z-roth:
@jamescd18 are there any pre-existing examples of the use of icons and tooltips that I can look at to implement those UI improvements?

RChandler234:
> @jamescd18 are there any pre-existing examples of the use of icons and tooltips that I can look at to implement those UI improvements?

There should be an example of tooltips on the implemented changes section of the change request detail page. Not sure about making the button an icon, I'd look up a react-bootstrap example or smthn

src/frontend/components/check-list/check-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+import { ReactNode, useState } from 'react';
+import PageBlock from '../../layouts/page-block/page-block';
+import { Form, Button, Modal } from 'react-bootstrap';
+import styles from './check-list.module.css';
+
+interface CheckListProps {
+  title: string;
+  headerRight?: ReactNode;
+  list: {
```
RChandler234, line 14: Could we have a more descriptive name for this prop, especially because it's asking for these custom jsx objects. I'm thinking something like listItems?

src/frontend/components/check-list/check-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+import { ReactNode, useState } from 'react';
+import PageBlock from '../../layouts/page-block/page-block';
+import { Form, Button, Modal } from 'react-bootstrap';
+import styles from './check-list.module.css';
+
+interface CheckListProps {
+  title: string;
+  headerRight?: ReactNode;
```
RChandler234, line 13: What is this prop for? I'm not seeing it used in the risk log file?

src/frontend/components/check-list/check-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <Modal.Footer>
+        <Button variant="danger" onClick={handleClose}>
+          Close
+        </Button>
+        <Button variant="success" onClick={handleClose}>
+          Save Changes
+        </Button>
+      </Modal.Footer>
+    </Modal>
+  );
```
RChandler234, line 86: This modal should arguably be in its own file, especially given that we're gonna have to add in the functionality to actually add and save the risk in the database later.

Separate question @jamescd18 @anthonybernardi do you think it'd be worth making a component for modals like this? Seems like it could be something we might end up using a lot.

src/frontend/components/check-list/check-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+      {check.resolved ? (
+        <Button variant="danger">Delete</Button>
+      ) : (
+        <Button variant="success">Convert to CR</Button>
+      )}
+    </div>
+  ));
+
+  builtList.push(
+    <div>
```
RChandler234, line 62: So for this whole section it looks like you're taking in checks that you've declared elsewhere and then using it to render all these react components. I would reformat this so that bulletlist is kinda a custom bit of React in this file. I believe
`const BulletList = (checkList: any) => { return ()}` should work and then you put all the code for the map inside a div inside the return statement (maybe wrap that in a div) and I think that'll work and then you should be able to just add the Add New Risk button below it in the top div rather than appending it to the list of jsx components as you've done here. 

I also suggest this method because it allows you to call bulletlist below like `<BulletList checkList = {checks} />` rather than `{builtList}`; I think this looks a lil cleaner/ makes it more clear what bulletlist is/ what it takes in.

Let me know if this suggestion is confusing btw, I know it's a lot. It's a lil easier to show in code, so I can hop in a call or smthn if need be.

src/frontend/pages/ProjectDetailPage/project-view-container/risk-log/risk-log.tsx
```diff
@@ -0,0 +1,16 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import CheckList from '../../../../components/check-list/check-list';
+
+interface RiskLogProps {
+  risks: { details: string; resolved: boolean }[];
```
RChandler234, line 9: It might be worth adding this risk type to the types folder in utils as it's probably going to be used in several places once we actually hook up the feature. Thoughts @jamescd18?

Also the feature looks great btw! All my comments are just knitpicky stuff.

src/frontend/components/check-list/check-list.tsx
z-roth: I tried to model this component to be modeled after the BulletList component, which also had it, so I figured I would add it in here too (I'm assuming you mean the headerRight prop?)

src/frontend/pages/ProjectDetailPage/project-view-container/risk-log/risk-log.tsx
anthonybernardi: definitely agree with this, there actually should be one already

src/frontend/components/check-list/check-list.tsx
anthonybernardi: not a bad idea, but will depend on how similar they all are

src/frontend/components/check-list/check-list.tsx
anthonybernardi: yeah lets keep it for the same reason bulletlist has it, just in case

anthonybernardi:
i'm good with this from a visual standpoint. I'll leave it to Reid for the nitty gritty of the code

RChandler234:
This looks a lot cleaner! There might be some stuff we wanna change/ add to the checklist component later, but for the purpose of this ticket this is awesome. Gonna go ahead and approve.

anthonybernardi (approved):
dub

z-roth:
lgtm !!

